PR #35, opened 2015-01-20
Fix matrix DFT parity inversion reported by Anand (WIP)

josePhoenix:


poppy/matrixDFT.py
```diff
@@ -302,22 +263,10 @@ def DFT_fftstyle_rect(pupil, nlamD, npix, inverse=False):
...
-    #else:
-        #expYV = expYV.T.copy()                          # matrix npixY * npupY
-        #t1 = np.dot(expYV, pupil)                       # matrix npixY * npupX?     dot prod combined last axis of expYV and first axis of pupil.
-        #t2 = np.dot(t1, expXU)                          # matrix npixY * npixX
-        #print expYV.shape, pupil.shape, expXU.shape
-
-    #print ""
-
-    #return  nlamD/(npup*npix) *   t2 * dx * dy
     # normalization here is almost certainly wrong:
```
mperrin: I wonder if it's still the case that "normalization here is almost certainly wrong"...  Whoops.

poppy/matrixDFT.py
```diff
@@ -445,34 +326,29 @@ def DFT_adjustible_rect(pupil, nlamD, npix, offset=(0.0,0.0), inverse=False, **k
...
-
+    tuples or lists, using the usual Pythonic order (Y, X)
 
     Parameters
     ----------
     pupil : array
         pupil array (n by n)
     nlamD : float or tuple
-        size of focal plane array, in units of lam/D
+        size of focal plane array, in units of lamda / D
```
mperrin: "lambda" is spelled wrong here. :-)

josePhoenix: Got it!

poppy/matrixDFT.py
josePhoenix: What is it supposed to normalize to?

mperrin: Conservation of energy across the FFT.   The total intensity in the output plane should be equal to that in the input plane. (Perfectly true for infinitely large arrays sampled continuously; approximately true for any finite sampling)

I looked through the test code to refresh myself on this. It looks like test_matrixDFT's test_MFT_flux_conservation and test_MFS_fluxconsv_all_types verifies this. The tests pass for all of the flavors of matrixDFT, so I think the comment is out of date and the normalization is correct.

mperrin: Hmmm, although I note that the tests are only for the forward direction transforms. For completeness' sake we should maybe extend that to also test the inverse direction (after sorting out the parity issue from Anand).

josePhoenix:
That turned into a rather large refactoring. Oops. :camel:

In the process, I uncovered bugs in the DFT_combined function's handling of rectangular inputs and outputs which I had to fix ( #36 ).

I'm afraid I also lumped the fix for the parity inversion in with the rest of it. It turned out to be a pair of missing negative signs when building the complex exponential matrices (compare [new](https://github.com/josePhoenix/poppy/blob/a6f9d02e619e441b8d308f22a6d7d2acae91f5f0/poppy/matrixDFT.py#L178-L179) to [old](https://github.com/josePhoenix/poppy/blob/c5142b651ea65a0ac1457db06ddcb2a6c1fba27e/poppy/matrixDFT.py#L129-L130)).

josePhoenix:
As a sanity check, I re-ran a [test case notebook](http://nbviewer.ipython.org/urls/gist.githubusercontent.com/josePhoenix/422caa57076fc8b29777/raw/5a0c61f416ceb9182fb0b8adbf0f416d36571ee0/fft_vs_sam.ipynb) I had made to see if the issue Anand reported affected WebbPSF.

The plots look identical to the linked version; the only thing that has changed is the logging output. (Speaking of which, I'm adding another commit to restore a missing space in the logging form string.)

PR #48, opened 2015-01-23
Fix attempting to assign to pupil_diam

josePhoenix:
`pupil_diam` is a computed property, but code in FITSOpticalElement was trying to assign to it.

(It's entirely possible that I introduced this through incautious merging. :disappointed: )


mperrin:
Actually this is probably my fault. I noticed yesterday in the midst of other work that OpticalElement and subclasses were actually old-style Python classes rather than new-style classes. This was entirely by accident and had gone unnoticed to now.   One of the differences is that old-style classes are less strict about assigning a new attribute on top of a property inherited from a parent class, but new-style classes forbid that. 

I fixed it and thought I had gotten everything working, but I must not have exercised this branch of code.  

Which incidentally implies this has to be missed by the test suite - i.e. we're not testing the FITSOpticalElement.__init__ at all?  Should fix that too...

josePhoenix:
It manifested as the whole test suite failing for me, but I've been juggling branches and adding `super(Foo, self).__init__`s around in POPPY so it's difficult to say definitively! I do think we're exercising `__init__` though.

PR #55, opened 2015-01-27
Allow examination of intermediate wavefront planes when propagating with SAM, multiprocessing, or both

josePhoenix:
Refactored `OpticalSystem.propagate_mono` / `SemiAnalyticCoronagraph.propagate_mono` and `OpticalSystem.calcPSF` to separate the concerns of weighting and combining simulated optical planes. `propagate_mono` now knows nothing about the wavelength weights, and `calcPSF` does the combining of both the final product and the intermediate planes. As part of this refactoring, calculations with multiprocessing gained the ability to return their intermediate optical planes.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -1182,7 +1182,8 @@ class OpticalSystem(object):
 
         if conf.enable_speed_tests:
             t_start = time.time()
-        if self.verbose: _log.info(" Propagating wavelength = %g meters %s")
+        if self.verbose:
+           _log.info(" Propagating wavelength = {0:g} meters".format(wavelength))
         wavefront = self.inputWavefront(wavelength)
 
         intermediate_wfs = []
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -1360,7 +1360,7 @@ class OpticalSystem(object):
                 if outFITS is None:
                     # for the first wavelength processed, set up the arrays where we accumulate the output
                     outFITS = mono_psf
-                    outFITS[0].data = mono_psf[0].data * wave_weight
+                    outFITS[0].data *= wave_weight
                     intermediate_wfs = mono_intermediate_wfs
                     for wavefront in intermediate_wfs:
                         wavefront *= wave_weight  # modifies Wavefront in-place
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -1235,7 +1235,7 @@ class OpticalSystem(object):
         return wavefront.asFITS(), intermediate_wfs
 
     def calcPSF(self, wavelength=1e-6, weight=None, save_intermediates=False, save_intermediates_what='all',
-                display=False, return_intermediates=False, source=None, normalize='first'):
+                display=False, return_intermediates=False, source=None, normalize='first', display_intermediates=False):
         """Calculate a PSF, either multi-wavelength or monochromatic.
 
         The wavelength coverage computed will be:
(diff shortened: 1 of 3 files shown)


mperrin:
Cool, this sounds really good. I will try to read through the diff in the next night or two sometime and see if I have any comments or suggestions.

poppy/poppy_core.py
```diff
@@ -1162,37 +1164,31 @@ def propagate_mono(self, wavelength=2e-6, normalize='first', save_intermediates=
...
+        intermediate_wfs : list
+            A list of `poppy.Wavefront` objects representing the wavefront at intermediate optical planes.
+            The 0th item is "before first optical plane", 1st is "after first plane and before second plane", and so on.
+            (n.b. This will be empty if `retain_intermediates` is False.)
         """
 
         if conf.enable_speed_tests:
             t_start = time.time()
-        if self.verbose: _log.info(" Propagating wavelength = %g meters %s" % (wavelength, "" if poly_weight is None else " with weight=%.2f" % poly_weight))
+        if self.verbose: _log.info(" Propagating wavelength = %g meters %s")
```
mperrin: This format statement appears to lack any formatting function call to expand the `%g` and `%s`?

poppy/poppy_core.py
```diff
@@ -1315,88 +1304,98 @@ def calcPSF(self, wavelength=1e-6, weight=None,
...
+                    retain_intermediates=retain_intermediates,
+                    display_intermediates=display_intermediates,
+                    normalize=normalize
+                )
 
                 if outFITS is None:
+                    # for the first wavelength processed, set up the arrays where we accumulate the output
                     outFITS = mono_psf
-                    outFITS[0].data = mono_psf[0].data*wave_weight
+                    outFITS[0].data = mono_psf[0].data * wave_weight
```
mperrin: this could be `outFITS[0].data *= wave_weight` for a minor savings, yes?

poppy/poppy_core.py
```diff
@@ -1315,88 +1304,98 @@ def calcPSF(self, wavelength=1e-6, weight=None,
...
+            # be sure to cast to int, will fail if given a float even if of integer value
+            pool = multiprocessing.Pool(int(nproc))
 
             # build a single iterable containing the required function arguments
             _log.info("Beginning multiprocessor job using {0} processes".format(nproc))
-            iterable = [(self, wavelen, wt, kwargs, _USE_FFTW) for wavelen, wt in zip(wavelength, normwts)]
-            results = pool.map(_wrap_propagate_for_multiprocessing, iterable)
+            worker_arguments = [(self, wlen, retain_intermediates, normalize, _USE_FFTW)
+                                for wlen in wavelength]
+            results = pool.map(_wrap_propagate_for_multiprocessing, worker_arguments)
```
mperrin: Since all the wavelengths and their intermediate wavefronts are accumulated here before any combining is done, this could potentially get very memory intensive for large numbers of wavelengths, many intermediate planes, etc. 

That's not necessarily a problem - and I don't see any easy work around - but we should make a note somewhere in the documentation, to the effect that users should be aware of this and should consider using single process mode if they run into memory limitations.

poppy/poppy_core.py
josePhoenix: Good catch! I fixed one instance and missed the other.

poppy/poppy_core.py
josePhoenix: :+1:

poppy/poppy_core.py
josePhoenix: :+1:

josePhoenix: Added a warning!

poppy/poppy_core.py
josePhoenix: Fixed!

poppy/poppy_core.py
josePhoenix: Fixed!

Issue #51, opened 2015-01-25
Set up continuous integration support for POPPY


josePhoenix:
Once our test suite is all passing and can run headless, we should set up [Travis-CI](https://travis-ci.org). Travis-CI will let us see when pull requests break things, as well as let users see if our test suite is passing on `master`.

mperrin:
Travis is working nicely. OK to close this one now.

PR #63, opened 2015-01-28
Enable Travis-CI for POPPY

josePhoenix:
I edited the `.travis.yml` template from Astropy to test only on NumPy 1.9 and 1.8, and to install our dependencies before running `egg_info` (the lack of which was killing the builds).

Now the build failures appear to actually be test failures (things that are trying to open plot windows and can't, for example).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: python
 
 python:
-    # - 2.6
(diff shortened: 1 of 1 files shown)


josePhoenix:
Addresses #51

PR #65, opened 2015-01-30
Fix for #64, added image normalization to display_profiles() and display_EE()

douglase:
See issue #64

### DIFF
--- a/poppy/utils.py
+++ b/poppy/utils.py
@@ -344,10 +344,11 @@ def display_EE(HDUlist_or_filename=None,ext=0, overplot=False, ax=None, mark_lev
         ax.set_ylabel("Encircled Energy")
 
     if mark_levels:
-        for level in [0.5, 0.8, 0.95]:
+        for fraction in [0.5, 0.8, 0.95]:
+            level=fraction*np.max(EE)
             EElev = radius[np.where(EE > level)[0][0]]
-            yoffset = 0 if level < 0.9 else -0.05 
-            plt.text(EElev+0.1, level+yoffset, 'EE=%2d%% at r=%.3f"' % (level*100, EElev))
+            yoffset = 0 if level < 0.9 else -0.05
+            plt.text(EElev+0.1, level+yoffset, 'EE=%2d%% at r=%.3f"' % (fraction*100, EElev))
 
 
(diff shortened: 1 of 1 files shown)


poppy/utils.py
```diff
@@ -330,8 +332,18 @@ def display_EE(HDUlist_or_filename=None,ext=0, overplot=False, ax=None, mark_lev
...
 
-    radius, profile, EE = radial_profile(HDUlist, EE=True)
+    if normalize.lower() == 'peak':
+        _log.debug("Displaying PSF normalized to peak = 1")
+        im /= im.max()
+    elif normalize.lower() =='total':
+        _log.debug("Displaying PSF normalized to PSF total = 1")
+        im /= im.sum()
+        
+    HDUlist_copy = fits.HDUList([fits.PrimaryHDU(data=im,header=HDUlist[ext].header)])
```
mperrin: This could be simplified using `deepcopy` from the standard library: 

```
import copy
HDUlist_copy = copy.deepcopy(HDUlist)
im = HDUlist[ext].data
```

poppy/utils.py
```diff
@@ -373,8 +388,18 @@ def display_profiles(HDUlist_or_filename=None,ext=0, overplot=False, title=None,
...
+    im = HDUlist[ext].data.copy() # don't change normalization of actual input array, work with a copy!
+
+    if normalize.lower() == 'peak':
+        _log.debug("Displaying PSF normalized to peak = 1")
+        im /= im.max()
+    elif normalize.lower() =='total':
+        _log.debug("Displaying PSF normalized to PSF total = 1")
+        im /= im.sum()
+        
+    HDUlist_copy = fits.HDUList([fits.PrimaryHDU(data=im,header=HDUlist[ext].header)])
```
mperrin: Hmm, the fact that the exact same code is needed in two places suggests maybe this normalization code belongs inside `radial_profile()` itself.  Then `display_EE` and `display_profiles` could both just pass the normalization parameter through as part of the kwargs.

poppy/utils.py
douglase: Yeah, that makes everything simpler. the commits below move the normalization to radial_profile().

PR #72, opened 2015-02-13
Test Gaussian jitter support in POPPY

josePhoenix:
n.b. this only tests that the FWHM increases and that the image peak ratio reported in the FITS header is not unity.

Using `FWHM_jitter = sqrt(FWHM_nojitter**2 + FWHM_jitterkernel**2)` overestimated the resulting FWHM. I'm not sure if this is because the kernel being used is actually the wrong size, because Airy disks are not sufficiently Gaussian-like, or because I've made an error in computing the FWHM for the kernel from the sigma value.

This also sneaks in a fix to the `AUTHOR` header entry, which was relying on environment variables that aren't necessarily available everywhere to set the `user@host` string.

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -1,6 +1,8 @@
 from __future__ import (absolute_import, division, print_function, unicode_literals)
 import os
 import time
+import platform
+import getpass
 import copy
 import numpy as np
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


mperrin:
The issue here with `FWHM_jitter = sqrt(FWHM_nojitter**2 + FWHM_jitterkernel**2) ` is just that there's a difference between the FWHM of a function and the sigma parameter of the Gaussian. And that difference is a factor of about 2.35.  The code looks like it's working fine.

I'll merge your PR and update the test to add in a revised test for the FWHM size being as expected.

mperrin:
Wait, I take that back - there actually is a bug here!  My test cases where it was working was just calling _applyJitter on an input array, not doing full PSF calculations. Ran into a problem after all when I tried to update your test case. See #77

PR #78, opened 2015-03-08
Add minimum versions (targeting Spring 2015 SSBrel) for setup

josePhoenix:
A bug in the Astropy helpers prevented automatic dependency resolution for `pip install poppy`. This was fixed in a patch release to Astropy, so we can avoid it by requiring `astropy>=1.0.1`. 

We should also update our copies of the various astropy package template and helper files.

See astropy/astropy#3541 for additional details.


josePhoenix:
Hmm. @embray, this build is failing to install astropy 1.0.1 on travis. using the dev version of astropy seems to work.

josePhoenix:
(It worked fine locally...)

josePhoenix:
Looks like both updated helpers and improved version constraints are needed. See #79

embray:
I haven't looked at your .travis.yml yet, but normally pip (for example) should just install whatever the latest version is by default.

embray:
Oh, I see.  It looks like the .travis.yml you're using will by default install astropy using conda as a prerequisite (so it normally always has it).  It's using the astropy-ci-extras channel on binstar (https://binstar.org/astropy-ci-extras) to get the astropy pacakge, I think, and that hasn't been updated to astropy 1.0.1 yet.  I don't maintain that (I think @astrofrog might?)  I think, if that's the case, that needs to be a regular part of our release procedure, which it currently isn't.

When you add `astropy >= 1.0.1` to your setup.py then that should force it to download and build the latest astropy.  But that shouldn't be necessary.  IIRC the point of using the binary packages from binstar is to speed up your builds (so it doesn't have to build Astropy every time just to run your own tests).  That seems preferable too, IMO.  (Maybe if you wanted to test installation on a bare environment you could have one build just for that too.)

astrofrog:
@embray - I've started the 1.0.1 packages to build: https://github.com/astropy/conda-builder/commit/ead6b717a2cf7a7557aa90c323dabf35bc2e8690

PR #79, opened 2015-03-08
Require minimum dependency versions, update Astropy helpers

josePhoenix:
This adds minimum version specifications for POPPY dependencies and an updated version of the Astropy helpers (v1.0.1).

To get the CI builds to pass with my changes, I had to put them into a single pull request. They're pretty closely related, anyway.


josePhoenix:
Aha, this might be just the ticket. All but one build task passes now. Going to relax the version requirement on numpy and try again.

PR #86, opened 2015-04-10
Python 3 compatibility

josePhoenix:
Really, very little needed to be done on our side.

What I did, with the help of `2to3`, was:

  * Convert `print` statements to `print()` calls
  * Add a dependency on `six` and use `six.string_types` where testing for string-y things
  * Make a version-agnostic way to test the contents of an exception message for our test cases that do that
  * Coerce some generators (e.g. from `range` or `zip`) into lists before passing them into `np.asarray` (so we don't end up with a 1-element array of a single generator object)


josePhoenix:
Looks like now that I've done the necessary tweaks by hand, the `astropy_helpers` exceptions have disappeared. Probably an artifact of some stuff being un-importable. I think this should be good to merge, after all!

mperrin:
Fantastic, thanks much.

PR #99, opened 2015-07-06
(Re-)Optimizes Zernike computation

josePhoenix:
Using the LRU cache decorator from Python 3.2+ (and a vendored copy in `poppy.vendor.lru_cache` for back-compatibility), cache individual Zernike polynomial arrays to speed ZernikeWFE computations.

This adds a test case that generates an input wavefront and compares calling `zernike1()` with `rho` and `theta` arrays (the way previous implementation did, but alas un-cacheable) to a new style where we pass in enough information to reproduce the rho and theta aways within `cached_zernike1()`.

~~The performance improvements ended up being "modest" (that is, ~2x), so there may be further productive profiling to be done.~~ Performance improved by a factor of ~6x for code that used zernike1 a lot.

An obvious next step would be to cache ZernikeWFE phasors with an internal instance attribute and some way to mark it dirty when `self.coefficients` or `self.radius` are assigned to (or make those attributes read-only).

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -826,56 +826,102 @@ class Wavefront(object):
         self.history.append('Rotated by %f degrees, CCW' %(angle))
 
 
+    @staticmethod
+    def pupil_coordinates(shape, pixelscale):
+        """Utility function to generate coordinates arrays for a pupil
+        plane wavefront
+
+        Parameters
+        ----------
+
+        shape : tuple of ints
+            Shape of the wavefront array
+        pixelscale : float or 2-tuple of floats
+            the pixel scale in meters/pixel, optionally different in
+            X and Y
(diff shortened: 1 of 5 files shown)


josePhoenix:
The WIP WFIRST field dependence stuff hits `zernike1` ~220 times in a monochromatic calculation, so it's a good realistic test-case to see if we made this any better.

 * In the uncached case this is **55 seconds** for a 10 wavelength calculation incorporating field dependent aberration.
 * All else being equal, caching makes the first calculation **~17 seconds**, and subsequent calculations are **~9 seconds**.

josePhoenix:
As discussed, refactored `Wavefront.coordinates` into static methods for image- and pupil-plane coordinates, then used those in `cached_zernike1`.

Issue #95, opened 2015-05-19
Implement Fresnel Propagation


douglase:
I've made a first attempt at extending POPPY to support Fresnel propagation in a branch in my fork: 

https://github.com/douglase/poppy/tree/fresnel

The new module inherits the Wavefront class and added the necessary functions, methods and properties for Fresnel diffraction calculations. I chose a separate submodule because don't want to break any existing POPPY routines. 

This new class has passed my initial tests. Demos of a telescope, diffraction from a circular aperture and the necessary Gaussian waist calculations are in this notebook: http://nbviewer.ipython.org/gist/douglase/600d6e4d88336f4ed0c4 (I left debug enabled, so there's lots of details in the notebook).

Some future work:
* Formal testing
* Make docstrings Astropy Guideline complaint 
* Streamlined logging

Is this a feature others would find useful? 

I'm going to keep working and testing this, but I wanted to open it up for outside input while things are still fresh.

mperrin:
Whoa. You have just totally made my day.  Yes this is absolutely a feature others will find useful! And it was on my own road map to figure out some time to implement in the next year or two.  Man this is fantastic. I promise to take a careful and close read through of this later, but am traveling for the next two weeks so will not be able to look at it in detail until likely June.   But this looks great just from skimming the notebook.

josePhoenix:
Very, very cool! :+1: 

I was looking through the code and thought I should mention some of the naming conventions used semi-consistently in POPPY (for ease of eventual integration with the main branch): class names are `CapitalizedLikeThis`, except for classes where the words are capitalized acronyms (e.g. `FQPM_FFT_Aligner`), and method names are `camelCase`.

(Blair Porterfield says hi, by the way!)

mperrin:
We should at some point think about cleaning up the main codebase to be more pep8 compliant. That's a relatively low priority overall and I admit I did not think much about it when I started writing all this in 2010 so there are parts that are messier than I would write today. Hard to be sure whether improving this is worth breaking back-compatibility; at some point we could add pep8 compliant names as synonyms and have a transitional period where both work.

douglase:
@mperrin I'm a little confused, do you feel extensions should be pep8, or follow the existing convention? 
I'm all over the map and still need to better absorb pep8.  

@josePhoenix say hello to Blair!

josePhoenix:
[...]

PR #100, opened 2015-08-02
Add fresnel module

douglase:
An implementation of Fresnel diffraction propagation. See also issue #95.


josePhoenix:
Thanks for this @douglase! I'll have a look through the code in the next week or two.

josePhoenix:
I was reading through the code to try and wrap my head around it, and ran into something odd which I haven't been totally able to pin down.

There's a set of magic constants in fresnel.py:
```
# internal constants for types of plane
_PUPIL = 1
_IMAGE = 2
_DETECTOR = 3 # specialized type of image plane.
_ROTATION = 4 # not a real optic, just a coordinate transform
_INTERMED = 5
```

Which is different from the values in `poppy_core.py`:
```
# internal constants for types of plane
_PUPIL = 'PUPIL'
_IMAGE = 'IMAGE'
_DETECTOR = 'DETECTOR' # specialized type of image plane.
_ROTATION = 'ROTATION' # not a real optic, just a coordinate transform
```

I took a stab at normalizing this by making them the same in both, but found that the Hubble demo in the notebook looks totally different (like it's plotting the wrong plane) when _PUPIL in fresnel.py is changed from `1` to `'PUPIL'`. Any ideas?

mperrin:
Those internal constants used to be ints originally, and I remember at some point one of us changed them to strings for slightly less inscrutability. Can't remember if I did that or you, @josePhoenix . I agree we need to have those consistent between poppy_core and fresnel...

josePhoenix:
My gut feeling on normalizing the planetypes:

  * remove the underscores, so we're clear these are for "public use", after we...
  * ...change `is _PUPIL` type tests to `== _PUPIL`, so passing in the value or the variable reference has the same effect.

I'm slightly in favor of going back to integer valued magic constants, as well. (If we were Python 3.x only, we could use https://docs.python.org/3/library/enum.html, but alas.)

mperrin:
The Py34 enum has been back ported (https://pypi.python.org/pypi/enum34) which admittedly would add another dependency for people on 2.7 but I'm OK with that. Yet another subtle factor to encourage people to upgrade to 3.4. :-)  I guess we should ask @cslocum first if adding a dependency on enum34 would be problematic for inclusion in Ureka though...

mperrin:
Oh and yes you are 100% right about switching to `==` instead of `is` for this comparison. Good catch!

josePhoenix:
Actually, if we use enums, the two are equivalent! I've got a branch that does that, but will clean it up a bit before PRing (and see if I can update the fresnel code to use it).

josePhoenix:
@douglase It seems like something in the `fresnel.py` code is sensitive to whether `_PUPIL` is an integer or not, which I can't figure out. (It's also possible that the bug is elsewhere in POPPY and we never saw it before.)

Here's a comparison, running your `fresnel.py` code and the HST demo calculation before and after making a one-line change to import `_PUPIL` from `poppy_core`. The plot of the amplitude and phase at the focus show something weird going on.

https://gist.github.com/josePhoenix/462096e25fff10267a75

douglase:
@josePhoenix,  greatings from SPIE.
 You definitely uncovered a bug. The planetype of the new analytic optical element, QuadPhase was hard coded as a pupil and that was overriding the user initialized 
intermediate planetype  of the secondary mirror.

Even though the planetype was wrong,  the example worked because fresnel.py used the old numerical planetype convention, and the check for if we were in a pupil plane was failing since the parent poppy. Wavefront class was initializing the planetype as u’PUPIL’, then when you fixed the planetype in fresnel propagate() thought it was in a pupil plane and dropped in the incident curvature.

I've fixed that, and made any Fresnel propagate wavefront intermediate unless the user specifies an optic is image or pupil.

This should fix the wavefront phase and amplitude when you use the poppy planetypes, but breaks display(), since the core plotting defaults to the image plane, the axes are no longer in meters. Not sure if we want to make display in poppy_core more general or just make a function that passes the wavefront to display as a pupil for plotting purposes?

josePhoenix:
@mperrin and I were just digging into this. Thanks!

Oh, and I defer to him on how to change the `display()` stuff. It *sounds* like that brushes up against what he was just describing to me as a consequence of using angular units for image planes, which was a design decision he understands the ramifications of better than I do!

mperrin:
Which display does it break, Wavefront.display() or OpticalElement.display?  

I haven't totally figured out the best overall API given that now there are two classes for regular-Wavefront and fresnel-Wavefront. (I sort of would like to rename the latter to FresnelWavefront in fact, to avoid having two different classes named Wavefront in the same package)

What I'd said before to @josePhoenix is somewhat related, which was that I wasn't sure if the Fresnel code would have trouble with IMAGE type optics, since those have dimensions in angular units. I was thinking of that in terms of propagation to optics and applying wavefront errors, not displays, but it's related of course.

mperrin:
meanwhile, some github machinery coordination: @josePhoenix already merged a copy of this pull request in, but as a separate branch `poppy/fresnel` rather than into `poppy/master`. This will enable us to collaborate on further debugging, while still keeping it distinct from the main `poppy/master` until we all are happy that it's ready to go for public release. 

Since that branch exists (and I see that @josePhoenix and I both just nearly-simultaneously merged your latest commits into it, whoops :-), your code is now effectively merged in. 

I think at this point it would be the least confusing if we were to close this PR without merging it into master, and then we can just do further development with PRs against the fresnel branch from the three of us?  Is that a reasonable plan? 

Further comments to be posted on issue #95

douglase:
Wavefront.display(), it didn't break it exactly, just labeled axes in arcseconds rather than meters. 

You are right about images planes being a problem, right now there is no handling of that case. 

Just got message regarding new branch, I can close this pull request.

PR #101, opened 2015-08-11
Implement PlaneType with an Enum

josePhoenix:
Introduces a dependency on the Python 3.4+ `enum` module, but the `enum34` backport is there as a dependency for older versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,8 +82,8 @@ install:
     # conda for packages available through conda, or pip for any other
     # packages. You should leave the `numpy=$NUMPY_VERSION` in the `conda`
     # install since this ensures Numpy does not get automatically upgraded.
-    - $CONDA_INSTALL numpy=$NUMPY_VERSION matplotlib scipy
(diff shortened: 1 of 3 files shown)


mperrin:
Before I merge this, I would just like to hear from @cslocum confirmation that it's OK to add a dependency to the `enum34` package on py27, and have that added in as part of Ureka as well. I figure this is almost certainly acceptable given the inclusion of other packages like `six` and `unittest2` that are back ports and compatibility layers for 2.7 vs 3.4. Agreed?

josePhoenix:
cc @jhunkeler since it sounds like you're currently working on packaging

mperrin:
Paging @cslocum and @jhunkeler . It's been a week, if I don't hear any objections from either of you within the next day or two to adding the dependency on `enum34` to `poppy` , I'm going to go ahead and merge this.

jhunkeler:
@josePhoenix pinged me via email. These github messages were going to my spam folder for some reason. Anyway, yes I'll add `enum24` to ssbdev tonight.

PR #105, opened 2015-08-17
Add a hook for inserting aberration optics after the pupil

josePhoenix:
Adds `poppy.Instrument.get_aberrations` which returns a `poppy.OpticalElement` or `None`. Used in the implementation of the WFIRST WFI model.

Additionally, lets POPPY select the units of OPD based on a FITS header keyword instead of forcing them to be microns.

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -474,7 +474,7 @@ class Instrument(object):
 
 
         #---- apply pupil intensity and OPD to the optical model
-        optsys.addPupil(name='Entrance Pupil', optic=pupil_optic, transmission=full_pupil_path, opd=full_opd_path, opdunits='micron', rotation=self._rotation)
+        optsys.addPupil(name='Entrance Pupil', optic=pupil_optic, transmission=full_pupil_path, opd=full_opd_path, rotation=self._rotation)
 
         # Allow instrument subclass to add field-dependent aberrations
         aberration_optic = self.get_aberrations()
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -476,6 +476,10 @@ class Instrument(object):
         #---- apply pupil intensity and OPD to the optical model
         optsys.addPupil(name='Entrance Pupil', optic=pupil_optic, transmission=full_pupil_path, opd=full_opd_path, opdunits='micron', rotation=self._rotation)
 
+        # Allow instrument subclass to add field-dependent aberrations
+        aberration_optic = self.get_aberrations()
+        if aberration_optic is not None:
+            optsys.addPupil(aberration_optic)
 
         #--- add the detector element. 
         if fov_pixels is None:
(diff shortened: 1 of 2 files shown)


poppy/poppy_core.py
```diff
@@ -2158,17 +2158,15 @@ def __init__(self, name="unnamed optic", transmission=None, opd= None, opdunits=
...
-
-            if opdunits.lower().endswith('s'): opdunits = opdunits[:-1] # drop trailing s if present
-            if opdunits.lower() == 'meter' or opdunits.lower() == 'm':
-                pass # no need to rescale
-            elif opdunits.lower() == 'micron' or opdunits.lower() == 'um' or opdunits.lower() == 'micrometer':
+            opdunits = opdunits.lower()
+            # rescale to meters if necessary
+            if opdunits in ('meter', 'meters', 'm'):
+                pass
+            elif opdunits in ('micron', 'microns', 'um', 'micrometer', 'micrometers'):
```
mperrin: what makes this preferable versus optionally dropping the trailing s if present, like in the prior code?

poppy/instrument.py
```diff
@@ -474,8 +474,12 @@ def _getOpticalSystem(self,fft_oversample=2, detector_oversample = None, fov_arc
 
 
         #---- apply pupil intensity and OPD to the optical model
-        optsys.addPupil(name='Entrance Pupil', optic=pupil_optic, transmission=full_pupil_path, opd=full_opd_path, opdunits='micron', rotation=self._rotation)
+        optsys.addPupil(name='Entrance Pupil', optic=pupil_optic, transmission=full_pupil_path, opd=full_opd_path, rotation=self._rotation)
```
mperrin: So does this require an immediate update to all the OPD files, or else it will raise the exception about missing units? Or does it handle legacy files semi gracefully in some way?

josePhoenix: Well, the two options are:

1. Require a webbpsf-data update with units on OPD maps, have the default OPD units be meters
2. Make the default OPD units microns, but do not require a webbpsf-data update

I thought the benefit to consistency (meters everywhere) was worth it, but we could change the default too

poppy/poppy_core.py
josePhoenix: Nothing, really. I find it more readable, and it saves calling `.lower()` 5 times, but that could be accomplished with a version that strips the 's' as well.

poppy/instrument.py
josePhoenix: Oh, actually, there isn't a fallback path. It has to be specified one way or another. But explicit is better than implicit :)

poppy/poppy_core.py
mperrin: OK, that's fine. This is micro-optimizing something that's not at all computationally demanding :-) I was mostly just idly curious since I couldn't figure out any technical need for this associated with the BUNIT changes.

josePhoenix: I wanted the code to look pretty... but I guess including the plural versions of each unit depresses the signal-to-noise as much as `.lower()`. Oh well.

poppy/instrument.py
mperrin: This would be a good case for incrementing the version number on the data files, then. 

I thought we had implemented some check for needing an updated version of that, but I don't see it anywhere in the code. There is a "version.txt" in the data directory but I don't see any code that checks that? I've got to be misremembering this somehow.

josePhoenix: Yep! I was thinking the same thing. I'll add that in to webbpsf.

josePhoenix:
Hmm, upon reflection this should probably be `_get_aberrations` rather than `get_aberrations.

mperrin:
This looks ready to go to me, and the test passes now. Just to confirm you're not planning to add anything else in to this PR? If so I can merge it now.

josePhoenix:
The only thing is if you think it's a good idea to have a fallback interpretation of OPD maps supplied without units. Since it would still break existing installs if we defaulted to meters, and it would be perhaps more surprising than just erroring out, I'm inclined to stick with having the code be pedantic about needing units supplied.

Aside from that, good to go!

mperrin:
I agree that erroring out is the least likely behavior to yield incorrect results. Explicit is better than implicit, errors should never pass silently, and all that.

Issue #106, opened 2015-08-17
Fresnel - Circular Aperture test case failing.


mperrin:
I just reimplemented the Anderson and Enmark test case in PROPER, and get a peak intensity=3.13422, which agrees within the available precision with the value 3.134 that I digitized from the Anderson and Enmark PDF figure using GraphClick. 

So that is clearly the value we should be aiming for. But not the value we are currently getting. :-(  

That disagrees with both the value I get with @douglase's original code (peak=3.36), or my current code after various edits related to #103 and #104 (peak=3.89).  In fact it looks like something is pretty off because the number of oscillations is quite different. See attached.  I will try to dig into this more.

![circularaperturepropagation](https://cloud.githubusercontent.com/assets/1151745/9313681/4224b674-44f2-11e5-942f-bd17d895bdc8.png)

mperrin:
Found it. in [`fresnel.Wavefront.ptp()`](https://github.com/mperrin/poppy/blob/fresnel/poppy/fresnel.py#L429) when calculating the `rho` value, the x and y separations are divided by 2. I believe that factor of 2 is unnecessary and wrong. 
Changing that gets the periodicity of the oscillations correct:  

![circularaperturepropagation_v2](https://cloud.githubusercontent.com/assets/1151745/9314448/080734ee-44f7-11e5-880f-925fb747bd84.png)

The peak value is still a bit high compared to our references, and the POPPY output is visibly more jagged than PROPER. I believe both of these may be due to aliasing from the discrete (0 or 1) pupil amplitude, compared with the smoother gray edge that PROPER draws circles with. I've tested this by increasing the sampling and the POPPY output maybe does trend toward the PROPER and Anderson & Enmark values with increasingly fine sampling (at a cost in speed). I'm not totally sure though.   

A while ago I started working (in a separate branch) to implement similar grayscale antialiasing of edges in POPPY, but never got that working entirely robustly. Might be worth revisiting that now, I suppose.  Though I kind of don't want to be juggling too many significant code overhauls at the same time!

mperrin:
> The peak value is still a bit high compared to our references, and the POPPY output is visibly more jagged than PROPER. I believe both of these may be due to aliasing from the discrete (0 or 1) pupil amplitude, compared with the smoother gray edge that PROPER draws circles with. I've tested this by increasing the sampling and the POPPY output maybe does trend toward the PROPER and Anderson & Enmark values with increasingly fine sampling (at a cost in speed). I'm not totally sure though.
[...]

PR #107, opened 2015-08-18
bringing direct_propagation() up-to-date

douglase:
for debugging/comparison with plane-to-plane, from working issue: #106

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -24,6 +24,9 @@ except:
 
 from poppy.poppy_core import _PUPIL, _IMAGE, _DETECTOR, _ROTATION, _INTERMED, _FFTW_AVAILABLE
 
+if _FFTW_AVAILABLE:
+    import pyfftw
+
 #conversions
 _RADIANStoARCSEC = 180.*60*60 / np.pi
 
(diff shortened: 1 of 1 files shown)


douglase:
Issue #106 identified an error in the plane-to-plane propagation. This function implements the direct propagation algorithm given in Andersen and Enmark, which was originally included for exactly this purpose of testing the POPPY implimentation of the plane-to-plane algorithm.

Example code snippet:

    import poppy
    from poppy import fresnel
    import astropy.units as u
    z=5000*u.m
    wf_direct=fresnel.Wavefront(2*u.m,wavelength=2200e-9,npix=1024,oversample=4)
    wf_direct *= poppy.CircularAperture(radius=0.5,oversample=wf_direct.oversample)
    wf_direct.propagate_direct(z)

gives a result that agrees well with @mperrin's bug fix and PROPER result:
circ_aperture_direct

mperrin:
Thanks. The above code would make a nice unit test, to run both the direct and PTP algorithms and check that the output result is consistent. 

In what situations do you think it is preferable to use the direct as opposed to the PTP algorithm? Is there a clear winner for speed or numerical accuracy, or anything like that? I am not that familiar with this direct algorithm.

PR #108, opened 2015-08-21
Fresnel pull request for commenting

mperrin:
This is the whole fresnel branch including my recent changes. PR for commenting as requested by @josePhoenix in https://github.com/mperrin/poppy/issues/103#issuecomment-133493924

In retrospect I probably should have made a branch from the branch and pull requested from that, but I didn't think of it.


poppy/poppy_core.py
```diff
@@ -34,18 +33,22 @@
     _FFTW_AVAILABLE = False
 
 # internal constants for types of plane
-_PUPIL = 'PUPIL'
-_IMAGE = 'IMAGE'
-_DETECTOR = 'DETECTOR' # specialized type of image plane.
-_ROTATION = 'ROTATION' # not a real optic, just a coordinate transform
-#_typestrs = {0:'', 1:'Pupil plane', 2:'Image plane', 3:'Detector', 4:'Rotation', 'PUPIL':'Pupil plane','IMAGE':'Image plane'}
+class PlaneType(enum.Enum):
+    unspecified = 0
```
josePhoenix: What's the rationale for this / where is it used? Would None work as a sentinel value instead?

poppy/poppy_core.py
```diff
@@ -437,16 +432,16 @@ def display(self,what='intensity', nrows=1,row=1,showpadding=False,imagecrop=Non
...
                 title = title.replace('after', 'after\n')
                 title = title.replace('before', 'before\n')
-            plt.title(title)
-            plt.xlabel(unit)
+            ax.set_title(title)
+            ax.set_xlabel(unit)
             if colorbar: plt.colorbar(ax.images[0], orientation='vertical', shrink=0.8)
 
-            if self.planetype ==_IMAGE:
+            if is_angular:
```
josePhoenix: Would it be possible for FresnelWavefront to accept the desired units in a subclassed version of `Wavefront.display`? If it's only used for plotting, I'm leery of adding more state to the wavefront

docs/overview.rst
```diff
@@ -115,16 +124,9 @@ An `~poppy.OpticalSystem` consists of a series of two or more planes, of various
...
-namely those which are well modeled by the Fraunhofer (far-field)
-approximation. If you need to model full physical propagation through
-intermediate optics, including Fresnel or Talbot effects, you should consider
-using another tool. (It is possible that POPPY will eventually be extended to
-include Fresnel propagation, but so far it has not yet become an immediate
-priority.)
-
-
-The plane type of a given optic is encoded by its `.planetype` attribute. 
+POPPY thus is capable of representing a moderate subset of optical imaging systems,
```
josePhoenix: Hah, nice, going from "limited" to "moderate" subset. :+1:

poppy/poppy_core.py
mperrin: I was trying to avoid subclassing `display` since it's a substantial hunk of code. But more fundamentally the `FresnelWavefront.angular_coordinates` attribute is *not* just used for display. Its main purpose is actually to select what coordinate system is used when computing the phasor; the display stuff is secondary.

In detail: If you do e.g. `somefresnelwave *= CircularOcculter()` then `CircularOcculter.getPhasor` expects that `somefresnelwave.coordinates()` returns coordinates in arcseconds.  The `angular_coordinates` attribute is what tells it to do so. 

In some sense you can think of `angular_coordinates` as an "act as if you were an image type wavefront" toggle. 

Under the Fresnel formalism, there is no real physical distinction between image and pupil planes; everything is just a plane somewhere in space with coordinates with meters. So even if it's convenient to think of them as "oh, this one is supposed to be an image", I decided it was best to be explicit that there is a coordinate toggle happening, rather than trying to do it always implicitly only based on inference from plane types. It still does that inference as best as possible, it just does so by setting an explicit and user-visible boolean.

josePhoenix: Ahhh, gotcha. Sounds good! :+1:

I'm going to fiddle with display() and see if there's a way to change the behavior from FresnelWavefront without copy-and-pasting code.

poppy/poppy_core.py
mperrin: None could work, yes. It's used in `FresnelWavefront.apply_lens_power()`, to let you stick optics in the system that don't affect or don't care what the plane type is.   Its effect is "pass the `wavefront` through this optic, leaving the `wavefront.planetype` unchanged". 

The original motivating use case is in the Hubble example:
 - Circular Aperture in a pupil plane
 - primary mirror (*which is still a pupil plane*)
 - secondary mirror (intermediate plane)
 - image plane

In the earlier version of this code (e.g. your https://gist.github.com/josePhoenix/462096e25fff10267a75) note that the lenses had to have a plane type explicitly assigned. In my refactored code you can just stick a lens there without specifying the plane and it works as desired. 

More broadly, all of the "pupil plane" optics from `optics.py` can potentially be stuck at intermediate planes as well as pupils, or even images now. E.g. the ACS coronagraph occulting spot is in the aberrated beam near but not precisely in focus, or the WFIRST PIAACMC design uses a series of Lyot stops at intermediate planes.  So I think it's useful to be able to say "just let the light go through this without relabeling what plane type the wavefront is at."

More fundamentally, the entire `planetype` paradigm is not a great fit for Fresnel because the boundaries are not sharp. It's kept for compatibility with the rest of poppy of course. But if we were to make a standalone Fresnel code I would advocate getting rid of the entire notion of `planetype` and just having wavefronts with a coordinates toggle between physical and angular units.

josePhoenix: Hmm, maybe renaming it to something like "inherit_from_previous" or similar would make its effects more transparent?

josePhoenix: (When I see 'undefined' I tend to think 'undefined behavior' but it sounds like the behavior is well-defined.)

poppy/tests/test_fresnel.py
```diff
@@ -0,0 +1,276 @@
...
+    expected_fwhm = 1.028*0.5e-6/2.4*206265
+    # we only require this to have < 5% accuracy with respect to the theoretical value
+    # given discrete pixelization etc.
+    assert(np.abs((measured_fwhm-expected_fwhm)/expected_fwhm) < 0.05)
+
+    ### check the various plane types are as expected, including toggling into angular coordinates
+    pt = poppy_core.PlaneType
+    assert(np.allclose([w.planetype for w in waves], [pt.pupil, pt.pupil, pt.intermediate, pt.image]))
+    assert(np.allclose([w.angular_coordinates for w in waves], [False, False, False, True]))
+    assert(np.allclose([w.spherical for w in waves], [False, True, True, False]))
```
josePhoenix: These three assertions fail for me, because (I think) `allclose` doesn't like boolean values. I've rewritten them with simple loops here: https://github.com/josePhoenix/poppy/commit/695162fa7d968ea9069b8d66f926c585e4175c1a

However, even in the case where it's not choking on the boolean values, I get an assertion failure on this line.

```
        spherical_flags = [False, True, True, False]
        for idx, (wavefront, spherical) in enumerate(zip(waves, angular_coordinates_flags)):
            assert wavefront.spherical == spherical, assert_message.format(
>               idx, "spherical", spherical, wavefront.spherical
            )
E           AssertionError: Expected FresnelWavefront at plane #1 to have spherical == False, but got True
```

mperrin: That's odd. It all works fine for me of course. What version of python are you using? On 2.7 with numpy 1.92 there is no problem with allclose: 
In [2]: np.allclose([True, False, True], np.asarray([True,False,True]))
Out[2]: True

mperrin: The lack of a spherical wavefront after the primary lens/mirror is somewhat more concerning of course. This should be getting set in `apply_lens_curvature`, I think.

josePhoenix: Ah, I was wrong. It's not boolean values, it's enum'ed (PlaneType) values.

```
>       assert(np.allclose([w.planetype for w in waves], [pt.pupil, pt.pupil, pt.intermediate, pt.image]))
[...]
E       TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

josePhoenix: Still not sure why I'm getting nonspherical wavefronts! Those two commits in #109 are the only way my copy has diverged from yours..

mperrin: Does the output PSF itself look correct? Hmm.

douglase:
The notebook looks great!
But it is failing for me because the pyfftw import has been missing since eb09a36e7dfff9e690d37eea1411fa6040e19e6d, I fixed this in #107.
I will test with pyfftw uninstalled for the moment.

poppy/tests/test_fresnel.py
douglase: I am getting the same assertion failure as @josePhoenix (and also the allclose() failure without his modifications), the hubble PSF looks fine, identical to your committed notebook.
Python Version:  2.7.10 |Anaconda 2.2.0 (x86_64)|
Numpy: 1.9.2
Scipy: 0.16.0
Matplotlib: 1.4.3
h5py: 2.4.0

mperrin:
@douglase I just merged #107 for you so you can try it with pyfftw now. 

Meanwhile I've managed to reproduce the error you're getting, although it's not totally clear to me why; on one computer it works fine in 2.7 but fails in 3.4, but on another computer I can get the ufunc casting error on 2.7. Investigating... but I'm not optimistic I will figure this out in the next 15 minutes before I have to split. :-)

douglase:
side note on notebook: GaussianLens() is Gaussian because it's spherical, so it is a "Gaussian Optics" approximation of a lens... SphericalLens would be clearer.

mperrin:
Or "QuadraticLens" maybe? The point is that it's a pure quadratic phase term, such that it converts a plane wave into a converging spherical wave or vice versa.  In hardware this would be achieved by a parabolic mirror (e.g. off axis paraboloid). Using a *real* spherical lens or mirror doesn't get you a perfectly converging beam, it gets you a converging beam with spherical aberration. As no astronomer should ever forget. :-) 

I can imagine a future version of this which has a subclass "ConicLens" that allows precise description of conic asphere optics (such as the actual mirrors of Hubble, or the optics of JWST which are an ellipse, hyperbola, and ellipse, etc).

douglase:
yes, spherical optics, not a spherical optic... I like QuadraticLens.

mperrin:
I am planning to go ahead and merge this into `master` pretty soon. Any additional comments or issues before I do so? 

I took a pass at cleaning up the `fresnel.py` codebase to be closer to PEP8 compliant; mostly minor formatting stuff. One thing I did not fix yet: the PEP8 inspector in PyCharm complains about the function names for `FresnelWavefront.z_R` and `FresnelWavefront.R_c` since those are not lowercase.  This is an instance where the common mathematical notation for these variables is not in keeping with Python's style conventions. We could rename these to e.g. `z_rayleigh` and `r_curvature`, and maybe that would actually be clearer (though more verbose). But it's also nice to maintain a close consistency with the mathematical notation. Opinions?

douglase:
I don't see any issues with with merging into master. I just created an issue for adding anti-aliased apertures (#129), since that seems to be the missing feature for head-to-head comparisons with PROPER, but that is not a just Fresnel issue. I have the PROPER example microscope working and see less than than 1% difference between POPPY and PROPER (https://github.com/douglase/poppy_example_notebooks/blob/master/Fresnel/Microscope_Example.ipynb). Have not had time for more examples (the TDEM coronagraph example is built but not well tested), but I don't think that is a reason to hold up merging with master. 

I think the mathematical notation makes the code a bit more readable than the PEP8 versions, but understand the desire for consistency.

mperrin:
Ok thanks much. Nice to see the microscope example!  After further reflection I think we can keep the shorter names, I'll just make them all lowercase for PEP8 compliance. And I think it's a little easier to remember all lowercase `z_r` and `r_c` as opposed to mixed case (oppositely for the two parameters!). That sound fair enough?

josePhoenix:
:+1: to lowercasing. I also thought about providing short aliases to the longer names but it's probably not worth the additional confusion.

josePhoenix:
Woohoo!

PR #109, opened 2015-08-21
Factor out the Fresnel-specific stuff from Wavefront.display a little bit

josePhoenix:
Now Wavefront.display doesn't need to know anything about attribute names on FresnelWavefront, pushing the responsibility down (up?) a level to FresnelWavefront.display to request angular units when appropriate.

(Pull requests within pull requests... First time I've done this, but didn't want to push indiscriminately to mperrin/poppy@fresnel!)

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -129,13 +129,16 @@ class GaussianLens(QuadPhase):
 
 
 class FresnelWavefront(Wavefront):
+    angular_coordinates = False
+    """Should coordinates be expressed in arcseconds instead of meters at the current plane? """
+
     @u.quantity_input(beam_radius=u.m)
     def __init__(self,
                  beam_radius,
                  units=u.m,
                  rayleigh_factor=2.0,
                  oversample=2,
-                 **kwds):
+                 **kwargs):
         '''
         Wavefront for Fresnel diffraction calculation.
 
@@ -148,8 +151,8 @@ class FresnelWavefront(Wavefront):
         beam_radius : astropy.Quantity of type length
             Radius of the illuminated beam at the initial optical plane.
             I.e. this would be the pupil aperture radius in an entrance pupil.
-        units :
-            astropy units of input parameters
+        units : astropy.units.Unit
+            Astropy units of input parameters
         rayleigh_factor:
             Threshold for considering a wave spherical.
         oversample : float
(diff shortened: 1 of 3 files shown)


poppy/poppy_core.py
```diff
@@ -351,13 +352,17 @@ def display(self,what='intensity', nrows=1,row=1,showpadding=False,imagecrop=Non
             Display colorbar
         ax : matplotlib Axes
             axes to display into
+        use_angular_coordinates : bool, optional
+            Should the axes be labeled in angular units of arcseconds?
+            This is used by FresnelWavefront, where non-angular
+            coordinates are possible everywhere. Behavior for Wavefront
+            is undefined.
```
mperrin: The comment 'Behavior for Wavefront is undefined' is sort of cryptic. I wonder if it could be phrased more clearly as some variant of 'For non-Fresnel Wavefront, you don't need to worry about this.'

I suppose adding a new function parameter is less of a kludge than checking for a subclass attribute, yes.

josePhoenix: Okay, I'll amend that docstring

poppy/fresnel.py
```diff
@@ -129,13 +129,16 @@ def __str__(self):
 
 
 class FresnelWavefront(Wavefront):
+    angular_coordinates = False
+    """Should coordinates be expressed in arcseconds instead of meters at the current plane? """
```
mperrin: Hmm, I generally try to avoid class attributes since they rise confusion and error if you use anything mutable there, and don't realize that it's shared among all class instances, but I suppose this will work properly with a bool.  Still I am not sure it's any better than assigning it in `__init__`...?

josePhoenix: Yeah, they're tricky, but I think this is a case where we should be using them.

1) It prevents the issue where pixelscale assignment in `Wavefront.__init__` tries to invoke the `FresnelWavefront.pixelscale` property setter and barfs because `angular_coordinates` is not yet set. This lets us move the super() call to the beginning, making the code easier to reason about (don't have to worry about whether things need to be assigned before or after calling the superclass initializer)
2) Sphinx (or one of the astropy extensions) can read docstrings for class attributes but not attributes defined in `__init__`.. This is perhaps minor
3) Someone would have to go out of their way to assign to `FresnelWavefront.angular_coordinates`; assigning to `wavefront_instance.angular_coordinates` will still do the right thing.

poppy/tests/test_fresnel.py
```diff
@@ -261,10 +261,24 @@ def test_fresnel_optical_system_Hubble(display=False):
...
+        )
+
+    angular_coordinates_flags = [False, False, False, True]
+    for idx, (wavefront, angular_coordinates) in enumerate(zip(waves, angular_coordinates_flags)):
+        assert wavefront.angular_coordinates == angular_coordinates, assert_message.format(
+            idx, "angular_coordinates", angular_coordinates, wavefront.angular_coordinates
+        )
+
+    spherical_flags = [False, True, True, False]
+    for idx, (wavefront, spherical) in enumerate(zip(waves, angular_coordinates_flags)):
```
mperrin: ah. Here you should be comparing spherical to `spherical_flags`, not to `angular_coordinates_flags`. :-)

josePhoenix: Oh heck. 😓

PR #110, opened 2015-08-26
Add the opd_expand function to poppy.zernike

josePhoenix:
This function (originally by @mperrin) expands a given wavefront OPD array in terms of a given basis set (Zernike by default). This PR also adds a test that verifies input coefficients for a generated OPD array are recovered within 0.1% when sampled at 512 pixels across the wavefront.

### DIFF
--- a/poppy/zernike.py
+++ b/poppy/zernike.py
@@ -308,8 +308,9 @@ def zernike_basis(nterms=15, npix=512, rho=None, theta=None, **kwargs):
         Desired pixel diameter for circular pupil. Only used if `rho`
         and `theta` are not provided.
     rho, theta : array_like
-        Image plane coordinates. `rho` should be 1.0 at the desired
-        pixel radius, `theta` should be in radians.
+        Image plane coordinates. `rho` should be 0 at the origin
+        and 1.0 at the edge of the circular pupil. `theta` should be
+        the angle in radians.
 
     Other parameters are passed through to `poppy.zernike.zernike`
     and are documented there.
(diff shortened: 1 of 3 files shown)


josePhoenix:
This now also removes the `mask_outside` keyword argument, and makes the `outside` argument required for user-supplied basis functions. It's unlikely anyone had implemented their own, so hopefully that's not a big change. Instead of supporting `masked_array` return values, the only "mask" is by assigning a value to the outside-rho-1 pixels. For OPD maps in ParameterizedWFE, we want `0.0`, but for `expand_opd` we want `np.nan` so we're not counting contributions from pixels outside the aperture. (Also, `plt.imshow` handles `nan` values gracefully when plotting computed Zernikes, so that's nice.)

poppy/zernike.py
```diff
@@ -309,33 +300,30 @@ def zernike_basis(nterms=15, npix=512, rho=None, theta=None, **kwargs):
...
+        and `theta` are not provided.
     rho, theta : array_like
-        Image plane coordinates. `rho` should be 1 at the desired pixel radius,
-        `theta` should be in radians
-    noll_normalize : bool
-        As defined in Noll et al. JOSA 1976, the Zernikes are normalized such that
-        the integral of Z[n, m] * Z[n, m] over the unit disk is pi exactly. To omit
-        the normalization constant, set this to False. Default is True.
+        Image plane coordinates. `rho` should be 1.0 at the desired
+        pixel radius, `theta` should be in radians.
```
mperrin: I think I didn't really write this clearly: "should be 1.0 at the desired pixel radius". The way you phrased it above on lines 200-202 for the `zernike` docstring is better, and I think it would make sense to copy that phrasing here?

mperrin:
This all looks good apart, from the minor issue with a docstring I pointed out above. Looks ready to merge to me pending that minor tweak.

josePhoenix:
All set!

PR #111, opened 2015-08-27
Handle a corner case in hex_aperture for even and odd npix

josePhoenix:
`hex_aperture` was behaving differently depending on even or odd npix values.

The best option we could come up with was picking only pixels entirely enclosed in the hexagon for inclusion in the aperture. Now, even and odd npix values alike will produce the behavior where the corners on the non-flat sides lack a pixel that would take them to the edge of the array. (In other words, hex_aperture arrays have a zero column or row between the corner and the edge of the array.)

### DIFF
--- a/poppy/tests/test_zernike.py
+++ b/poppy/tests/test_zernike.py
@@ -136,3 +136,28 @@ def test_cross_hexikes():
     """
     for testj in (2, 3, 4, 5, 6):
         _test_cross_hexikes(testj=testj, nterms=6)
+
+def test_hex_aperture_even_odd_npix():
+    """Issue reported by Alex Greenbaum where even-npix
+    hex apertures had an extra zero column on each side
+    but odd-npix ones did not.
+    
+    In an effort to make the behavior consistent, the `hex_aperture`
+    function now only includes pixels entirely enclosed by the hexagon
+    in the aperture, so the pointiest corner near the array edge will
+    always be chopped off (not included) resulting in a zero column
+    at the edge.
+    """
+    from poppy.zernike import hex_aperture
+
+    odd_hex = hex_aperture(npix=127)
+    assert np.max(odd_hex[:,0]) == 0.0
+    assert np.max(odd_hex[:,1]) == 1.0
+    assert np.max(odd_hex[:,-2]) == 1.0
+    assert np.max(odd_hex[:,-1]) == 0.0
+
+    even_hex = hex_aperture(npix=128)
+    assert np.max(even_hex[:,0]) == 0.0
+    assert np.max(even_hex[:,1]) == 1.0
+    assert np.max(even_hex[:,-2]) == 1.0
+    assert np.max(even_hex[:,-1]) == 0.0
--- a/poppy/zernike.py
+++ b/poppy/zernike.py
@@ -359,7 +359,12 @@ def hex_aperture(npix=1024, rho=None, theta=None, vertical=False):
         the whole array from edge to edge in the direction aligned
         with its flat sides. (Ignored when `rho` and `theta` are
         supplied.)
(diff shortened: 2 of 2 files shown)


mperrin:
Can you explain a bit more the motivation / use case behind this one? Under what circumstances does the presence or absence of zero rows or columns matter in this way?

I'm both trying to understand more clearly exCtly what trouble the current code was causing, and also wondering whether this needs to be handled consistently across other aperture shapes. For instance I don't think you will get zero rows always for circular or square apertures. 

What would be the correct behavior in the case of having subpixel antialiased values instead?

josePhoenix:
Alex Greenbaum pointed out that the behavior was inconsistent depending on the parity of npix, in that sometimes (for odd npix values greater than about 63) there would be 1.0 values in columns 0 and -1, and other times (for even npix values at all sizes) those two columns would be all zeroes.

The best I could do fiddling with the inequalities in hex_aperture was make the behavior consistently like the even-npix case, but after sleeping on it I thought of another way to rearrange them which might ensure they reach the edge (instead of using less than or equal to, use strictly greater than and invert the sense of the mask to identify pixels entirely outside).

I'm not actually sure what issues Alex had with it that prompted her report.

josePhoenix:
> What would be the correct behavior in the case of having subpixel antialiased values instead?

I think it would be closer to the behavior in this PR, since the existing behavior for odd npix values is essentially "rounding up" on corner pixels that should some shade of gray.

mperrin:
For the record, I have been ignoring / leaving aside this PR for now, because I still don't entirely understand the root issue or the desired behavior. So I am not sure how to assess if this fix is the right one, nor whether other apertures would need similar fixes. 

Perhaps this will all jus the obviated once we have properly gray scaled apertures?

josePhoenix:
Basically it's a question of "should pixels not entirely within the aperture be marked as having 100% transmission", from what I remember. The current behavior (I think) erroneously marks some corner pixels as being entirely within the hex aperture

mperrin:
Oh, somehow I missed that this was about `zernike.hex_aperture` as opposed to `optics.HexagonAperture`. 

I've just done some testing and confirmed that those two functions do *not* give consistent outputs. The hexagon is often about 1 pixel smaller as returned from `hex_aperture`. The definitions of the hexagon aperture in terms of the rectangle and left and right triangles are identical between the two; both use `<=` for the comparisons. 

However it turns out they are using different coordinate calculations for the pixel `x,y` values:  `poppy.Wavefront` (and by extension all optical element classes) takes the diameter as the total distance across the array: if you have a 1 meter radius and 10 pixels, then it is 0.2 meter per pixel, and the outermost pixel centers are at 0.9 m from the center of the array (i.e. the pixel extends across 0.8 - 1.0 meters from the center). `hex_aperture` takes the radius (implicitly 1) as the distance to the outer pixel centers of the array, so the outer pixels would be exactly at 1.0 pixels.  Each pixel is then 0.222 repeating meters across. 

So I think the fix here is going to be to make the coordinates inside `hex_aperture` become consistent with the convention used by `poppy.Wavefront.image_coordinates`

mperrin:
Addressed by 36395f0 but I (again) put the wrong issue number in the commit message.

mperrin:
I think the fix I just committed in  36395f0 is the proper way of addressing this. Alex can you take a look? Partially it has the opposite effect of Joseph's changes - in that the first  column of pixels will now always have a nonzero pixel rather than always a zero pixel - but the end result is the same, more consistent output from the function. And now there is a test case that enforces consistency between `hex_aperture` and `HexagonAperture` with hexagon side length=1. 

Given that I am inclined to set aside the code changes in this PR. Does that sound OK?

PR #113, opened 2015-09-02
Fresnel - minor units fix and allow optics to work on non-_PUPIL wavefronts

douglase:
In working on test cases, discovered the assertion test in most optics get_phasor prevents you from using them on intermediate wavefronts, to fix changes the test to not apply the optic to an image, rather than requiring a pupil.


mperrin:
Yeah excellent - I had realized this was going to be an issue, just hadn't done anything about it yet. Thanks again!

PR #115, opened 2015-09-17
Douglase quadphase shifted

### LINKED ISSUE #114: Fresnel - Propagation Bug
In implementing a validation coronagraph (based on the TDEM report 1, https://github.com/kuravih/poppy_example_notebooks/blob/TDEM_coronagraph/Fresnel/Coronagraph/TDEM_Coronagraph.ipynb) and the microscope prescription from the PROPER manual (https://github.com/douglase/poppy_example_notebooks/blob/master/Fresnel/Microscope_Example.ipynb), we found a significant error that crops up dramatically for small apertures and that shows this code is not quite ready for release.

for example, for a simple 1 lens system:

    beam_diameter =  .025*u.m
    fl = 0.7*u.m
    wavefront = fresnel.FresnelWavefront(beam_radius=beam_diameter/2., wavelength=500.0e-9, npix=256, oversample=2)
    diam = beam_diameter
    wavefront.apply_lens_power(fresnel.GaussianLens(fl, name='M1'))
    wavefront *= poppy.CircularAperture(radius=diam.value/2)
    wavefront.display("both")
    wavefront.propagate_fresnel(fl*2,display_intermed=True)
    wavefront.display("both")

 we see a non-circularly symmetric diffraction pattern at the ROC, independent of the oversampling:
propagation_bug_illustration17sept2015

This is a dramatic case of slighter asymmetries that have been visible in other tests, which I have not worried too much about, partly thinking they came from the hard edges of the apertures. But I've tried using a ant-aliased PROPER generated entrance aperture as the input, and while the higher-order effects go away, the quadrant pattern does not.

This seems to only effect the to and from spherical functions, since the plane-to-plane comparisons to PROPER show no such asymmetries. (#106).

I have implemented the test system above in PROPER and Zemax and both are circularly symmetric as expected.

I have limited time to debug this now but will look into it more when time allows.

douglase:
fixed #114.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -85,73 +85,21 @@ class QuadPhase(AnalyticOpticalElement):
         lens_phasor = np.exp(1.j * k * rsqd/(2.0*self.z_m))
         return lens_phasor
 
-class _QuadPhaseShifted(AnalyticOpticalElement):
-    '''
-    Quadratic phase factor,  q(z)
-    suitable for representing a radially-dependent wavefront curvature.
-    This function for operating on FFT shifted arrays, with the origin in the corner,
-     for centered "physical" coordinate system optics with an origin at the image center use 'QuadPhase'.
-
-    Parameters
-    -----------------
-    z : float or astropy.Quantity of type length
-        radius of curvature
-    planetype : poppy.PlaneType constant
-        plane type
-    name : string
-        Descriptive string name
-    reference_wavelength : float
-        wavelength
-    units : astropy.unit of type length
-        Unit to apply to reference wavelength; default is meter
-
-
-    References
-    -------------------
-    Lawrence eq. 88
-
-    '''
-    def __init__(self,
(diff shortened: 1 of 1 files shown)


mperrin:
Aha, the issue is needing to fftshift the phasor so it matches the FFTed data? Good catch!  I am actually kind of amazed it already worked as well as it did, given that error. 

That said I wonder if we can figure out a more elegant way to set up the fix - adding this class is a lot of duplicated code for what ends up being a one-line change. Here is a suggestion:

```
class _QuadPhaseShifted(QuadPhase):
    def getPhasor(self, wave):
        """ Return complex phasor, for FFT shifted array"""
        return np.fft.fftshift( super(_QuadPhaseShifted, self).getPhasor(wave)
```

Does that suffice?

douglase:
that works and is certainly more pythonic than my cut and paste. implemented above, thanks!
I think that it appeared to be working well just shows the Fresnel effects in the Hubble PSF are mostly negligible since the apertures are so large.

PR #117, opened 2015-10-15
Allow configurable colormaps and make image origin position consistent

josePhoenix:
In order to let us say things about orientation of features in POPPY images and be understood across different possible user configurations, we need to specify `origin='lower'` every time we plot an image. (We could alternatively change every call to `imshow` to use a wrapper in POPPY, or use matplotlib's context manager to swap in our rcParams values.)

Also, this might be a good time to use more-highly-recommended colormaps. I added config options for `cmap_sequential`, `cmap_diverging`, and `cmap_mask`; defaulting to `afmhot`, `RdBu_r` (reversed so higher values are red, lower are blue), and `gray` respectively.

## [**The old origin and cmap behavior** (notebook)](http://nbviewer.ipython.org/gist/josePhoenix/8d01c37e9c491b8c2c32)

Example:

old_cmap

## [**The new behavior** (notebook)](http://nbviewer.ipython.org/gist/josePhoenix/fe39f5451f2eb801c82c)

Example:

new_cmap

### DIFF
--- a/poppy/__init__.py
+++ b/poppy/__init__.py
@@ -73,7 +73,7 @@ class Conf(_config.ConfigNamespace):
         'verbose printout of fluxes and flux conservation during '+
         'calculations. Useful for testing.')
     cmap_sequential = _config.ConfigItem(
-        'gist_heat',
+        'afmhot',
         'Select a default colormap to represent sequential data (e.g. intensity)'
     )
     cmap_diverging = _config.ConfigItem(
--- a/poppy/__init__.py
+++ b/poppy/__init__.py
@@ -3,13 +3,13 @@
 """Physical Optics Propagation in PYthon (POPPY)
 
 
-POPPY is a Python package that simulates physical optical propagation including diffraction. 
+POPPY is a Python package that simulates physical optical propagation including diffraction.
 It implements a flexible framework for modeling Fraunhofer (far-field) diffraction
 and point spread function formation, particularly in the context of astronomical telescopes.
-POPPY was developed as part of a simulation package for JWST, but is more broadly applicable to many kinds of 
-imaging simulations. 
+POPPY was developed as part of a simulation package for JWST, but is more broadly applicable to many kinds of
+imaging simulations.
 
-Developed by Marshall Perrin at STScI, 2010-2014, for use simulating the James Webb Space Telescope. 
+Developed by Marshall Perrin at STScI, 2010-2014, for use simulating the James Webb Space Telescope.
 
 Documentation can be found online at https://pythonhosted.org/poppy/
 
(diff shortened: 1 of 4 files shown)


josePhoenix:
Hmm. I can't quite decide between gist_heat and afmhot for PSF plots. afmhot washes out the details in the WFIRST PSF cross-check plots because ~2 orders of magnitude are packed in the white-to-very-light-yellow region.

I'm thinking `gist_heat` was better for log scaled intensity images after all.

josePhoenix:
I noticed an unrelated issue where the `ext` keyword argument was not used in `display_profiles`, so I did the one line change to fix that as well.

mperrin:
In terms of selecting color maps - the plain gray colormap is actually quite good for seeing faint details. I am not suggesting using that as the default necessarily (I agree color is more visually satisfying) but comparing the amount of detail that can be seen in a gray plot can be a good standard yardstick against which to compare the afmhot and gist_heat. 'bone' and 'cubehelix' may also be worth looking at.

josePhoenix:
Good idea. Also, if we did use gray across the board, we'd only have one config item to set a sequential colormap, which feels better to me. (And one can always pass `cmap=` to `display_PSF`, if necessary!)

mperrin:
Looking over my shoulder here at this PR suddenly spawned a large and animated discussion of color map selection among the wavefront team here at JSC.  :-)   

The consensus was that it's nice to have separate color maps for PSFs versus pupil intensities; Alden Jurling and Scott Smith and I all favor keeping different versions for the PSFs and intermediate pupils rather than just putting everything all in gray. 

For sequential colormaps Scott likes 'pink', Alden uses 'gray' or 'hot'. But hot doesn't look monotonic in intensity to me so I would avoid that one. Everybody thinks that gist_heat looks pretty good.  For diverging, Alden suggested 'coolwarm' instead of 'RdBu_r', but I am unconvinced.  Or 'bwr'. 

Everybody has an opinion about color tables. :-)

josePhoenix:
All the more reason to make it configurable!

josePhoenix:
Just for fun, some combos of the suggestions. I'm going to see how coolwarm looks in this one WFIRST report figure that I haven't been able to make look quite as nice as I'd like; it might in fact be superior to RdBu_r.

josePhoenix:
### bone and bwr

bone_bwr

### pink and coolwarm

pink_coolwarm

josePhoenix:
Ack, no.

### RdBu_r

rdbu_r

### coolwarm

coolwarm

mperrin:
Overall I like this but before I merge it, I'm not totally sure about `cmap_mask` as the setting name. It's not just used for aperture masks, it looks like it's also being used for the intermediate pupil intensity in your example above? Should this be something like `cmap_pupil_intensity` instead?

josePhoenix:
Good point, that name makes more sense. I'll change it

mperrin:
speaking of which did you ever hear back anything from Erik about trying to un-break the Travis stuff?  I am baffled on that and don't really have time to dig into it, but now I'm really missing that with all these PRs to check...

josePhoenix:
I did not, alas. I dug into it as much as I had time to, but got effectively nowhere. I was going to dust off my branch without the astropy helpers and see if I could get that running.

PR #118, opened 2015-10-26
Run Instrument._validateConfig before trying to display the Instrument

josePhoenix:
This is mainly so auto_pupil type features have a chance to swap out the pupil shape. This may introduce unnecessary complexity.

There's also some trailing whitespace cleanup included.

### DIFF
--- a/poppy/instrument.py
+++ b/poppy/instrument.py
@@ -27,12 +27,12 @@ from . import utils
 __all__ = ['Instrument']
 
 class Instrument(object):
-    """ A generic astronomical instrument, composed of 
+    """ A generic astronomical instrument, composed of
         (1) an optical system implemented using POPPY, optionally with several configurations such as
             selectable image plane or pupil plane stops, and
-        (2) some defined spectral bandpass(es) such as selectable filters, implemented using pysynphot. 
+        (2) some defined spectral bandpass(es) such as selectable filters, implemented using pysynphot.
 
-    This provides the capability to model both the optical and spectral responses of a given system. 
+    This provides the capability to model both the optical and spectral responses of a given system.
     PSFs may be calculated for given source
     spectral energy distributions and output as FITS files, with substantial flexibility.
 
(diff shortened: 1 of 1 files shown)


mperrin:
can you explain more about the "this may introduce unnecessary complexity"?

josePhoenix:
I'm not happy with the way calls to "_validateConfig" are sneaking in all over, but I don't have a better idea yet. In theory every option that could affect the OpticalSystem should be a property that runs some validation on setting, making validateConfig unnecessary for getOpticalSystem/display/calcPSF, but that seems like a fairly large change and should maybe wait

josePhoenix:
If you look at the test case for the WFI pupil selection you'll see what I mean... Triggering validateConfig has to be treated specially in the test case, which feels awkward

mperrin:
The other thing I am not sure about is, the motivation/requirement for config validation is not entirely clear. For instance there are cases where you do want to ignore the automatic setting stuff, like in test cases.  `_validateConfig` was originally going to be e.g. "warning, you're using the MIRI coronagraph with a filter you can't actually use with that".  And the auto-config stuff kind of snuck in. 

Would the API make sense if there were separate functions? I'd be open to suggestions if you can think of some way to clarify things better. 

But I'm going to go ahead and merge this now, since you mentioned your WFIRST PR to webbpsf depends on this.

josePhoenix:
Hm, would the properties-with-validation-on-assignment strategy work for MIRI's coronagraph / filter combos? If there is a (coronagraph, filter) combo that *only* works with each other, you could end up with un-settable properties (e.g. trying to set the filter demands you set the coronagraph first, trying to set the coronagraph demands you change the filter)

josePhoenix:
I think if the only thing we do on incompatible configurations is emit a warning through the log system, it'd still work

mperrin:
Yes, the need to set e.g. `miri.image_mask='FQPM1140'` and `miri.filter='F1140C'` simultaneously was one of the reasons I didn't pursue implementing validation any more vigorously.  It seemed like this consistency checking was more trouble than it was worth.  You're right it would be OK to emit warnings though. 

You could enforce this via automatic property settings, though, with the logic being "oh, I see you are setting `image_mask="MASK1140"`, therefore I will automatically set the filter to F1140C for you". But this behavior seems like it might also violate the principle of least surprise if users are not expecting it.

josePhoenix:
Another thought: for non-interactive use, `WARN` level messages would make the signal-to-noise for the log output worse. Every time you switch optics you'd get a warning, even if you immediately set the corresponding filter. I think I see the argument for a separate validation step better now.

PR #121, opened 2015-10-30
Add ZernikeWFE.getOPD

josePhoenix:
Add hook to get OPD from ZernikeWFE before it's applied. If this proves to be useful, it could be pushed down to ParameterizedWFE or similar.


mperrin:
Ah, yes. I've thought about adding exactly this functionality before. More generally this idea is potentially applicable to any optical element. It would be a large refactoring to make all the AnalyticOpticalElements have a getOPD, and then getPhasor would become some small wrapper to that for the complex exponential. That's a longer term change but we should sketch out what a good API might look like.

douglase:
@mperrin, yes! getting the OPD in meters for the any optic would be incredibly useful, I was wishing, not for the first time, that I had that functionality just last week.

josePhoenix:
Does this API look good to you @douglase? Any other related "nice to have"s?

douglase:
That looks good to me.

mperrin:
Well, except in general an optic is defined by two 2D real quantities: an amplitude transmission/throughput and an OPD. If we have a `getOPD` then should there also be a `get_transmission`? 

One could imagine defining in `AnalyticOpticalElement` as follows: 
```
    def get_OPD(self, wave, units='meters'):
        return 0

    def get_transmission(self, wave):
        return 1
    
    def getPhasor(self, wave):
        opd_as_phase = 2 * np.pi * self.get_OPD(wave, units='meters')  / wave.wavelength
        phasor = self.get_transmission(wave) * np.exp(1.j * opd_as_phase)
        return phasor
```
Then subclasses could define their own `get_OPD` and `get_transmission` instead of `getPhasor` as currently.  Would that be a worthwhile approach? On the other hand this might lead to some repetition of code between the `get_OPD` and `get_transmission` functions, for instance setting up coordinate systems.  The other approach would be something like:

```
    def get_OPD_and_transmission(self, wave, units='meters'):
        return 0, 1

    def getPhasor(self, wave):
        opd, transmission = self.get_OPD_and_transmission(wave, units='meters')
        opd_as_phase = 2 * np.pi * opd  / wave.wavelength
        phasor = transmission * np.exp(1.j * opd_as_phase)
        return phasor
```

mperrin:
Oh and incidentally when adding new functions, should we strive for PEP8 compliance with underscores? That isn't something I kept in mind when starting this code base 5 years ago but we can be more mindful now perhaps.

josePhoenix:
I was trying to go with the 'house style' for now, and planning to do a Big Renaming Pull Request for 0.4. I just want to land the WFIRST stuff first, so right now there are a bunch of smallish PRs for POPPY that perhaps point the way to more refactoring once that has gotten released.

josePhoenix:
I like factoring out `get_phasor` into `get_opd` and `get_transmission` (where most subclasses won't have to touch `get_phasor` itself). Any non-trivial duplicated code there could be factored out into a private method, I think.

josePhoenix:
Whoops, I killed it. Thought I could resurrect Travis by merging the code that skips that one test, but git had other ideas.

PR #122, opened 2015-11-02
Fix FFTW wisdom loading on Python 3.x

josePhoenix:
FFTW wisdom was being loaded and saved without explicit encoding/decoding of strings, which is no longer valid in Python 3.x. I fixed that, made the (de)serialization use JSON so we don't have to worry about parsing the wisdom file (just putting stuff in and getting it back out again), and added a simple load/save test to the suite.


josePhoenix:
Oh and some weird thing with the astropy_helpers went away when I upgraded it, so that commit's in here too.

PR #125, opened 2015-11-03
Add ZernikeWFE.getOPD to get the OPD pre-wrapping in units of waves or meters

josePhoenix:


mperrin:
Here's a link to https://github.com/mperrin/poppy/pull/121 which this is a replacement for.

mperrin:
So did we decide to leave this as `getOPD` or use `get_OPD` to make it slightly more PEP8-compliant?  I agree leaving the mass renaming/refactoring for later but if we are adding a new function here, might as well be forward thinking?

josePhoenix:
Well, PEP8 would be `get_opd`. If you're okay with it, I'm all for preemptively PEP8ing this one function

mperrin:
well I wasn't quite sure how the PEP8 rules work with acronyms that naturally are all caps. Is that explicitly called out that it has to be all lowercase? I'm not opposed to it, it just (still) seems more natural to have acronyms as upper case to me.

josePhoenix:
The style guide doesn't mention acronyms specifically, but you can see examples of lowercased acronyms in a lot of places (e.g. [np.fft.fft](http://docs.scipy.org/doc/numpy/reference/generated/numpy.fft.fft.html#numpy.fft.fft), [urllib](https://docs.python.org/3.5/library/urllib.request.html#module-urllib.request))

josePhoenix:
For an example of acronyms in a CamelCase identifier (classes): http://matplotlib.org/api/backend_pdf_api.html#matplotlib.backends.backend_pdf.FigureCanvasPdf

I'm not sure if I think `FqpmFftAligner` is better than `FQPM_FFT_aligner` (both rankle somewhat), but I think there's more flexibility there than with capital letters in function names. It's very unusual to have both underscores and capitalization delimiting words in function / method names, in my experience.

mperrin:
good point re `np.fft`. OK let's go with `get_opd`.   But yuck I do not like `FqpmFftAligner`!

mperrin:
2 arc second seeing here at Gemini South, btw. Which is why I am working on this rather than observing anything.

Issue #102, opened 2015-08-11
Update .travis.yml for new travis container stuff


mperrin:
See http://docs.travis-ci.com/user/migrating-from-legacy/

This looks relatively straightforward. @cslocum is this something you could help with? Or anyone else in SSB who has already figured out the details of this type of transition. 

Incidentally at some point it would be great if someone more well versed in coveralls.io than me could take a look at the poppy and webbpsf coveralls config too, and figure out what I did wrong.

mperrin:
I've done this, and it's running on the new container infrastructure. 

But it's still failing with that inexplicable 247 error.

josePhoenix:
My `remove_helpers` branch fails with a more interesting error now: https://travis-ci.org/josePhoenix/poppy/jobs/88923255

It looks like this is to do with exhausting memory on the build server... I'm surprised, since we didn't have this before switching to the container infrastructure.

josePhoenix:
https://duckduckgo.com/?q=travis+error+137&t=ffsb

josePhoenix:
Evidently we have some very greedy tests in `test_poppy_core`... >1GB usage while those tests run on my laptop.

josePhoenix:
Fun trivia: the shell got to 137 by adding 128 to the kill signal (-9). http://unix.stackexchange.com/a/99134

josePhoenix:
`poppy.tests.test_core.test_optic_resizing` hits a high-water mark of 4 GB :exclamation:

@mperrin any ideas? It doesn't look like it's doing anything egregious... 1000 x 1000 x 64 bits is on the order of 8 MB for each wavefront array kicking around

mperrin:
Ah but the interpolation happens using `scipy.ndimage.interpolation.zoom` by default using cubic interpolation. So, who knows what's going on inside that black box.   I agree 4 GB seems egregious.

mperrin:
It's the very last line in that test function:

```
 tmp2 = test_optic_large_element.getPhasor(osys.inputWavefront())
poppy     : DEBUG    Creating input wavefront with wavelength=0.000002, npix=1024, pixel scale=0.006348 meters/pixel
poppy     : DEBUG    Pixelscales: wave 0.006348, optic 0.100000
poppy     : DEBUG    resampled optic to match wavefront via spline interpolation by a zoom factor of 15.8
poppy     : DEBUG    trimmed a border of 7365 x 7365 pixels from optic to match the wavefront
```

[...]

PR #126, opened 2015-11-03
Better rescaling test

mperrin:
A revised test for `OpticalElement.getPhasor` rescaling that avoids the overly-large arrays encountered in #102 
Also adds explicit tests for padding and cropping the resized arrays.


mperrin:
bah, I have totally failed to get the right commits into this. it's weird trying to set up a PR entirely within one's own repository.

PR #128, opened 2015-11-05
Added a MatrixFTCoronagraph subclass of OpticalSystem

neilzim:
This enables more efficient WFIRST SPLC propagation, and could benefit other coronagraph designs when the semi-analytical / Babinet superposition approach is not valid.


mperrin:
Nice. This makes sense as an addition and the code looks good. 

Could you put together some example code for using this? It would be good to have a test case that has some simple coronagraph for which this is suitable, and which shows that you get the same result (or close enough) by using the regular brute force FFT approach and by using `MatrixFTCoronagraph`.

neilzim:
Good idea - would definitely make sense to provide a simple example of this. The issue is, I don't have a ready-made SPLC design constructed with built-in (analytic) optic functions, as opposed to linking to some mask files. I don't mind sharing some masks publicly, but then the demo is no long self-contained within the code package. It would need to be a ring-type SPLC design, since those are defined based on a set of concentric annuli, which the optics.py functions already can make.

mperrin:
Does it have to be an SPLC? Couldn't we use something really basic like a simple classical Lyot coronagraph with a finite field stop at the intermediate image plane?

neilzim:
True, much easier to implement and it would still be a solid demo.

josePhoenix:
Related to the demo, this should probably be accompanied by a test case. (POPPY uses py.test, see the "test_*.py" files in poppy/tests/.) It doesn't have to be rigorous like doing an analytic solution and cross checking or anything, just some test of agreement with existing code so we get alerted if the two diverge (and so it is regularly exercised by our CI process)

mperrin:
Hi Neil, 
FYI we are likely to be making a new release version of poppy within the next two weeks (no later than end of November). I think there is no rush on this `MatrixFTCoronagraph` class so I am inclined to leave it until a future release - but if you want to put together a test case and example code in the next week or so, we could include this in the upcoming release.

mperrin:
Hi @neilzim, 

I wanted to get back to this PR before it got too stale. I'm happy to merge this but I do want to make sure we have at least one automated test case that exercises this code.  Your example notebook has all the ingredients for this. Could you work with @josePhoenix to make a unit test (in `poppy/tests/test_core.py`)? After that I think we will be ready to merge this in to master.

josePhoenix:
I believe this is in 7704168, so all that's necessary is moving it into `test_core`. (And, of course, verifying that it passes for all of us!)

mperrin:
bah, how did I miss that? My bad.  I'm fine with leaving it in `test_matrixdftcoronagraph`. Never mind then, looks like this is good to merge. :-)

mperrin:
test passes for me!

I'll probably tweak it a little to make the oversampling factor adjustable (and the tolerance threshold along with it), since right now this adds about 40% to the total test suite run time for poppy. But that's a minor thing and I can do it after this is merged. @josePhoenix do you think you can verify that the test also passes for you today, and then I'll go ahead and merge?

neilzim:
Sorry I missed the earlier part of this convo. Glad the test script checks out for you.

josePhoenix:
Test case passes for me!

PR #134, opened 2015-11-20
implementation of missing lens operator test

douglase:


mperrin:
aha, just in under the deadline for the 0.4 release, nice!


...hey, weren't you supposed to be launching a rocket right around now?  ah, Google tells me it's set for this coming Saturday after a weather scrub?  Good luck. We're rooting for you all. :-)  I am somewhat terrified of what it's going to feel like, 3 days before JWST launches...

mperrin:
shoot, I merged this into master but I forgot to pull it locally before making the 0.4.0 release (which I just did a little while ago). So this test is not included in the 0.4.0 package on pypi. We'll get it in the next release before too long.

Git is sometimes too complex for its own good.

Issue #142, opened 2016-01-27
better document the zernikeWFE class in poppy docs


mperrin:
@douglase tells me he couldn't find anything about the zernike optics class in the docs, and indeed it seems like the ZernikeWFE class is only mentioned in the release notes but not actually documented anywhere. 

@josePhoenix could you put together a couple paragraphs about this, and/or some notebook code?

josePhoenix:
For anyone searching for this, there is an example notebook already: https://github.com/mperrin/poppy/blob/master/notebooks/ThinLens_ZernikeWFE_and_ParameterizedWFE.ipynb

I'll write a more complete document for the online HTML docs.

mperrin:
Hi @josePhoenix can you give me an estimate for when you might be able to work on this? This and #144 are the only remaining issues for the next release.

josePhoenix:
Working on it right now, but got sidetracked by some refactoring in `wfe.py`

josePhoenix:
One thing is that I think I'll need to expose ZernikeWFE and friends under the main poppy namespace to get the `automodapi` helper to find them. This won't break any existing code.

mperrin:
Yes go for it. Exposing them in the main namespace also helps make them more apparent to users so it's win-win.

josePhoenix:
It appears that `hexike_basis` no longer works as described in the demo notebooks (or at all), so I'm investigating that as part of this issue

PR #151, opened 2016-02-10
Add documentation on wavefront error related classes

josePhoenix:
Addresses #142


josePhoenix:
Note this may fail CI because of some code changes necessary to make Sphinx build without warnings

docs/wfe.rst
```diff
@@ -0,0 +1,216 @@
+Representing sources of wavefront error
+=======================================
+
+When used for Fraunhofer-domain propagation, POPPY supports expressing aberrations as if they were applied in a pupil plane. Currently, there is the `~poppy.ZernikeWFE` optical element to represent wavefront error as a combination of Zernike terms, and the `~poppy.ParameterizedWFE` optical element that offers the ability to specify basis sets other than Zernikes.
```
mperrin: The first sentence should be rephrased - these are equally applicable in a Fresnel propagation as in Fraunhofer. There's no restriction on pupil vs image plane either. Zernikes are a basis set for parameterizing wavefront error over a circular aperture on some surface, regardless of what optical conjugate that surface might be at.   

E.g. it's totally correct to parameterize wavefront error over JWST's secondary mirror with a Zernike basis, even though that mirror is not at a pupil conjugate.

docs/wfe.rst
```diff
@@ -0,0 +1,216 @@
...
+
+ZernikeWFE
+----------
+
+For use in POPPY calculations, the `~poppy.ZernikeWFE` analytic optical element provides a way to introduce wavefront error as a combination of Zernike terms. The Zernike terms are both ordered in and normalized in the convention of Noll 1976. [#noll1976]_ This means that the polynomial terms carry a normalization factor to make :math:`\int_0^{2\pi} \int_0^1 Z_j^2\,\rho\,d\rho\,d\theta = \pi.` Additionally, the standard deviation of the optical path difference over the disk for any given term will be 1.0 meters. (See :ref:`zernike_normalization` for details on how this might impact your calculations.)
```
mperrin: First phrase 'For use in POPPY calculations,' seems redundant/unnecessary. Brevity is the soul of wit, etc.

docs/wfe.rst
josePhoenix: Re: fresnel: noted.

Re: pupil vs image: that's currently how the ZernikeWFE (and WavefrontError) class works. It supplies a plane type at instantiation. That might be something to change...

docs/wfe.rst
```diff
@@ -0,0 +1,216 @@
...
+       poppy.display_PSF(psf, title="PSF #{}".format(idx))
+
+.. figure:: ./figures/wfe/zernikewfe_wfe_budget.png
+
+   PSF of a 2.0 meter diameter circular aperture with randomized combinations of aberrations applied.
+
+.. _zernike_normalization:
+
+Understanding the normalization employed in POPPY
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
mperrin: Perhaps rephrase the title as "normalizing to desired peak-to-valley WFE" or something like that? Because this section is really about how to get a desired amount of peak-to-valley. 

You might consider adding a complementary (shorter) section on "normalizing to desired RMS WFE". That ought to make it very clear to users how to get whichever of those two they desired for their application.

docs/wfe.rst
```diff
@@ -0,0 +1,216 @@
...
+   osys.addPupil(thinlens)
+   osys.addDetector(pixelscale=PIXSCALE, fov_arcsec=FOV)
+
+   psf_with_zernikewfe = osys.calcPSF(wavelength=WAVELENGTH, display_intermediates=True)
+
+The resulting PSF, plotted with `poppy.display_PSF`:
+
+.. figure:: ./figures/wfe/zernikewfe_astigmatism.png
+   :scale: 50%
+
```
mperrin: Might be nice to also show an image of the corresponding OPD here too.

docs/wfe.rst
```diff
@@ -0,0 +1,216 @@
...
+       osys.addPupil(thinlens)
+       osys.addDetector(pixelscale=PIXSCALE, fov_arcsec=FOV)
+
+       psf = osys.calcPSF(wavelength=WAVELENGTH, display=False)
+       poppy.display_PSF(psf, title="PSF #{}".format(idx))
+
+.. figure:: ./figures/wfe/zernikewfe_wfe_budget.png
+
+   PSF of a 2.0 meter diameter circular aperture with randomized combinations of aberrations applied.
+
```
mperrin: Same comment as above - could show the OPDs as well as the PSFs. (would need to take care generate both figures from the same script to ensure consistency of the random coefficients)

docs/wfe.rst
mperrin: Pupil plane is fine as a default, but the point is the math should work just as well for any plane which has dimensions in meters available (e.g. anything other than a Fraunhofer-domain image plane calculation). 

Something to consider for eventual API larger overhaul: In retrospect I don't think that I should have made `planetype` a property of an optic itself. Rather it is a property of the ordering of planes that makes up an optical system.   The `OpticalSystem` should maintain the list of plane types along with the list of plane objects themselves. (And the list of distances between planes, in the case of the Fresnel.)

docs/wfe.rst
josePhoenix: Ah yeah, that was the beginning of the document at one point. This is why another pair of eyes is helpful :)

docs/wfe.rst
josePhoenix: I think it doesn't make sense to add a new section for RMS WFE, since the other examples are all implicitly about achieving some RMS WFE in various terms. I'll update the title of the normalization section and reword some of the other places I talk about coefficients to try and clarify that.

PR #152, opened 2016-02-10
Zernike bugfixes

josePhoenix:
While writing up the ZernikeWFE class, I found some refactoring opportunities in the `wfe.py` file and some bugs to fix in `zernike.py`.


poppy/wfe.py
```diff
@@ -20,16 +21,72 @@
 
 __all__ = ['WavefrontError', 'ParameterizedWFE', 'ZernikeWFE']
 
+def _accept_wavefront_or_meters(f):
+    """Decorator that ensures the first positional method argument
+    is a poppy.Wavefront or a floating point number of meters
+    for a wavelength
+    """
+
```
mperrin: Can you remind me what the motivation is for allowing either Wavefront or floating point wavelength as the first argument?  I guess it's just a convenience because it's often relevant to generate Zernikes over an abstract pupil without defining a spatial scale. 

I'm mostly wondering if this decorator is only relevant here in `wfe.py` or if there are conceivably other use cases such that it should be promoted to `optics.py`

poppy/wfe.py
```diff
@@ -20,16 +21,72 @@
...
+    Analytic optical elements that represent wavefront error should
+    derive from this class and override methods appropriately.
+    Defined to be a pupil-plane optic.
+    """
     def __init__(self, **kwargs):
         super(WavefrontError, self).__init__(planetype=_PUPIL, **kwargs)
 
+    @_accept_wavefront_or_meters
+    def get_opd(self, wave, units='meters'):
+        """Construct the optical path difference array for a wavefront error source
```
mperrin: This reminds me, I have a branch which has refactored most of `optics.py` to use this same approach, `get_opd`/`get_phasor`/ instead of the current `getPhasor` only. I should push that sometime, but I was hoping to leave that for after this release. 

Totally fine to include this here for now, of course. I think the generic `getPhasor` here will be redundant with the one on base `AnalyticOpticalElement` once that other branch is merged, but we can clean this up later on.

PR #162, opened 2016-04-21
GetPhasor -> get_opd, get_transmission refactor

mperrin:
This PR implements the API change proposed in #158 : the `getPhasor` function is replaced with `get_phasor` (renamed for PEP8 compliance), which then calls new functions `get_opd` and `get_transmission` which do the actual work.

The benefits of this are:
- Easier to define and extend subclasses since you can deal with the amplitude transmission and path delay components separately. Many optics such as pupil stops and coronagraph masks etc only need to define `get_transmission`, and so forth. 
- For plotting purposes it will let the actual OPD be displayed on screen even if the OPD is > 1 wave, which would get phase-wrapped before with the `getPhasor` implementation. 

Since this is a good sized API change, I'd like to get comments before merging it. @josePhoenix @neilzim @douglase what do you think?   Note that I've left in back-compatibility aliases for `getPhasor` for now, which can be removed in some future version. 

(note the commit history attached got a little screwy with repeated commits due to some git branch/rebase shenanigans. mea culpa. so, looking at the overall file deltas is a better way to see what changes than the individual commits)

PR #165, opened 2016-04-25
Use astropy.Quantities for all wavelengths everywhere

mperrin:
This addresses #145 by making lots of things into astropy Quantities, primarily wavelengths and pixel scales. As a result we can now explicitly specify wavelengths in meters or microns or whatever and things will Just Work. 

As proposed in #145, a decorator is added to ``utils`` that allows back-compatible function arguments given as floating point values to be automatically converted to astropy Quantity lengths in meters. You can give other units explicitly as e.g. ``wavelength=2*u.micron``.   (This decorator is currently named "quantity_input", the same as the related but different function provided by astropy itself. I couldn't think of any better name apart from monstrosities like e.g. ``@back_compatible_quantity_input_with_slightly_different_behavior_than_the_astropy_one``; please feel free to suggest improvements)

In any case, this decorator is then applied liberally throughout poppy_core, fresnel, and optics.py. Many string formatting and FITS header writing code lines had to be fixed. I added various casts to floating point values in meters where needed, which is awkward and a bit convoluted but is needed given that not absolutely everything uses Quantities yet.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -9,7 +9,6 @@ import time
 import poppy
 from poppy.poppy_core import PlaneType, _FFTW_AVAILABLE, OpticalSystem, Wavefront
 from . import utils
-from .utils import quantity_input
 
 _log = logging.getLogger('poppy')
 
@@ -47,15 +46,7 @@ class QuadPhase(poppy.optics.AnalyticOpticalElement):
                  name='Quadratic Wavefront Curvature Operator',
                  **kwargs):
         poppy.AnalyticOpticalElement.__init__(self, name=name, planetype=planetype, **kwargs)
-        self.z = z
-
-        #if isinstance(z, u.quantity.Quantity):
-            #self.z_m = z.to(u.m)  # convert to meters.
-        #else:
-            #_log.debug("Assuming meters, phase {:.3g} has no units for Optic: " .format(z) + self.name)
-            #self.z_m = z * u.m
-        self.z_m = z.to(u.m)  # convert to meters.
-
+        self.z = z.to(u.m)
 
     def getPhasor(self, wave):
         """ return complex phasor for the quadratic phase
@@ -68,12 +59,12 @@ class QuadPhase(poppy.optics.AnalyticOpticalElement):
 
         y, x = wave.coordinates()
         rsqd = (x ** 2 + y ** 2) * u.m ** 2
-        _log.debug("Applying spherical phase curvature ={0:0.2e}".format(self.z_m))
-        _log.debug("Applying spherical lens phase ={0:0.2e}".format(1.0 / self.z_m))
+        _log.debug("Applying spherical phase curvature ={0:0.2e}".format(self.z))
+        _log.debug("Applying spherical lens phase ={0:0.2e}".format(1.0 / self.z))
         _log.debug("max_rsqd ={0:0.2e}".format(np.max(rsqd)))
 
         k = 2 * np.pi / wave.wavelength
-        lens_phasor = np.exp(1.j * k * rsqd / (2.0 * self.z_m))
+        lens_phasor = np.exp(1.j * k * rsqd / (2.0 * self.z))
         return lens_phasor
 
 
@@ -127,11 +118,6 @@ class QuadraticLens(QuadPhase):
                            planetype=planetype,
                            name=name,
                            **kwargs)
-        #if isinstance(f_lens, u.quantity.Quantity):
-            #self.fl = f_lens.to(u.m)  # convert to meters.
-        #else:
-            #_log.warn("Assuming meters, focal length ({:.3g}) has no units for Optic: ".format(f_lens) + self.name)
-            #self.fl = f_lens * u.m
         self.fl = f_lens.to(u.m)
         _log.debug("Initialized: " + self.name + ", fl ={0:0.2e}".format(self.fl))
 
@@ -518,7 +504,7 @@ class FresnelWavefront(Wavefront):
 
     # methods for optical propagation
 
-    @quantity_input(z=u.meter)
+    @utils.quantity_input(z=u.meter)
     def propagate_direct(self, z):
         """
         Implements the direct propagation algorithm described in Andersen & Enmark (2011). Works best for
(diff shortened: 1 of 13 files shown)


poppy/fresnel.py
```diff
@@ -39,19 +40,21 @@ class QuadPhase(poppy.optics.AnalyticOpticalElement):
...
                  **kwargs):
         poppy.AnalyticOpticalElement.__init__(self, name=name, planetype=planetype, **kwargs)
         self.z = z
 
-        if isinstance(z, u.quantity.Quantity):
-            self.z_m = z.to(u.m)  # convert to meters.
-        else:
-            _log.debug("Assuming meters, phase {:.3g} has no units for Optic: " .format(z) + self.name)
-            self.z_m = z * u.m
+        #if isinstance(z, u.quantity.Quantity):
```
josePhoenix: -1 to keeping commented code blocks; if we need them we can always get them back from the revision history

poppy/fresnel.py
```diff
@@ -635,7 +639,7 @@ def _propagate_ptp(self, dz):
         x, y = self.coordinates()  # meters
         rhosqr = np.fft.fftshift(
             (x / self.pixelscale / self.oversample) ** 2 + (y / self.pixelscale / self.oversample) ** 2)
-        t = -1.0j * np.pi * self.wavelength * (
+        t = -1.0j * np.pi * (self.wavelength.to(u.meter).value) * (
```
josePhoenix: I don't think these parens are strictly necessary

poppy/fresnel.py
```diff
@@ -682,12 +687,14 @@ def _propagate_wts(self, dz):
         else:
             self._inv_fft()
 
-        self.pixelscale = self.wavelength * np.abs(dz.value) / (self.n * self.pixelscale)
+        #FIXME with units:
```
josePhoenix: Hmm, is this fixed already?

poppy/fresnel.py
mperrin: whoops, yeah, this wasn't intentionally - just not entirely cleaned up code from the (many) intermediate edits on the way towards getting this working. Good catch.

poppy/fresnel.py
```diff
@@ -713,17 +720,19 @@ def _propagate_stw(self, dz):
...
-        if dz > 0:
+        if dz > 0*u.meter:
             self._fft()
         else:
             self._inv_fft()
 
         # update to new pixel scale before applying curvature
-        self.pixelscale = self.wavelength * np.abs(dz.value) / (self.n * self.pixelscale)
+        #FIXME with units:
+        old_pixelscale = self.pixelscale
```
josePhoenix: I see a FIXME so I assume `old_pixelscale` will be eliminated.. just in case you missed it, it's not used anywhere

poppy/fresnel.py
mperrin: Although, ha, actually this whole part of the code can be simplified dramatically since before we were carrying both a `self.z` which was a float and a  `self.z_m` which was a Quantity in meters.  There's probably no longer any reason to do that.

poppy/optics.py
```diff
@@ -81,14 +83,15 @@ def __str__(self):
     def getPhasor(self, wave):
         raise NotImplementedError("getPhasor must be supplied by a derived subclass")
 
-    def sample(self, wavelength=2e-6, npix=512, grid_size=None, what='amplitude',
+    @quantity_input(wavelength=u.meter)
```
josePhoenix: I liked that `quantity_input` was namespaced under `utils.quantity_input` in `fresnel.py`. Might keep people from thinking it's the exact same as the astropy one... Other naming idea: `optional_units_input` or `optional_units_args`?

poppy/poppy_core.py
```diff
@@ -119,29 +120,30 @@ class Wavefront(object):
 
     """
 
-    def __init__(self,wavelength=2e-6, npix=1024, dtype=np.complex128, diam=8.0, oversample=2, pixelscale=None):
+    @quantity_input(wavelength=u.meter, diam=u.meter, pixelscale=u.arcsec/u.pixel)
+    def __init__(self,wavelength=2e-6*u.meter, npix=1024, dtype=np.complex128, diam=8.0*u.meter, oversample=2, pixelscale=None):
 
-        if wavelength > 1e-4:
-            raise ValueError("The specified wavelength %f is implausibly large. Remember to specify the desired wavelength in *meters*." % wavelength)
+        #if wavelength > 1e-3*u.m:
```
josePhoenix: I guess this is how POPPY becomes a general-purpose physical optics library? 😉 More commented code to be removed

poppy/fresnel.py
mperrin: yes, that was temporary for debugging - thanks for the careful reading.

poppy/poppy_core.py
```diff
@@ -1408,13 +1420,16 @@ def calcPSF(self, wavelength=1e-6, weight=None, save_intermediates=False, save_i
...
-            if np.isscalar(wavelength):
-                wavelength = np.asarray([wavelength], dtype=float)
-            else: wavelength = np.asarray(wavelength, dtype=float)
-        except (ValueError,TypeError):
-            raise ValueError("You have specified an invalid wavelength to calcPSF: "+str(wavelength))
+        # ensure wavelength is a quantity which is iterable:
+        # (the check for a quantity of type length is applied in the decorator)
+        if np.isscalar(wavelength.value):
+            wavelength = np.asarray([wavelength.value], dtype=float)*wavelength.unit
+        #try:
```
josePhoenix: Commented code

mperrin:
Kinda pleased that all the webbpsf unit tests pass against this with no units-related changes. Guess I got the back-compatibility part right!   ( #162 on the other hand did require some minor webbpsf changes for the custom optic classes there.)

And in the course of putting more units into the OpticalElement classes, I found and fixed two long-standing bugs we'd missed until now, with the NgonAperture class and the unit test for MFT forward & inverse transforms cancelling out. So that's a plus. 

I will probably go ahead and merge this relatively soon, and then shift development attention more towards webbpsf.

PR #172, opened 2016-07-08
Fix ConfigParser import that broke in 3.x

josePhoenix:
I'm not sure why we haven't encountered this before, but I just ran into it when installing POPPY in a fresh/clean conda Python 3.5 environment. (It might be a 3.4 to 3.5 change?)

(Fixed upstream in astropy/package-template#172)


josePhoenix:
Ah, shoot, committed some notebook changes as well. I'll try again.

PR #174, opened 2016-07-12
Fix #171: improper formatting of unit-y wavelength quantities

### LINKED ISSUE #171: issue with unit_check_wrapper
First of all: thanks for this great toolbox !

Unfortunately I have encountered an error with the unit_check_wrapper in poppy 0.5.0. If I try to run the unit tests or the example scripts the following error (or similar) occurs:

"Only dimensionless scalar quantities can be converted to Python scalars" or 
" 'm' (length) and '' (dimensionless) are not convertible"

I've tried this on a fresh installed system with Python 2.7 x64, Python 3.5 x64 and Python 3.5 x32 (Anaconda distributions). The same for the included ipython notebook files.

With poppy 0.4.1 the error doesn't occur.

What am I doing wrong ? Thanks in advance.

Sven

josePhoenix:
I did a pass through looking for places where wavelength was being suspiciously multiplied by numerical constants and caught another instance that needs fixing.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -1030,8 +1030,10 @@ class FresnelOpticalSystem(OpticalSystem):
         inwave = FresnelWavefront(self.pupil_diameter / 2, wavelength=wavelength,
                                   npix=self.npix, oversample=oversample)
         _log.debug(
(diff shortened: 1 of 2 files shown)


mperrin:
looks good, thanks.

PR #178, opened 2016-08-11
Python 3 pickle/json compatibility fixes

mmechtley:
Fixes issue [fftw_save_wisdom fails on python3 (during json serialization)](https://github.com/mperrin/poppy/issues/177)

Force pickle protocol=0 for python 2/3 compatibility
Use .decode('ascii') and encode('ascii') when (un)pickling for python 3 compatibility (pickle.dumps gives bytes, but json expects a string)


josePhoenix:
The tests appear to be passing on both 2.7 and 3.x, so I'm going to merge this. Thanks!

josePhoenix:
cc @kvangorkom and/or @kjbrooks I think you mentioned encountering this issue

PR #183, opened 2016-09-08
Wavefront class display() method: fix the handling of user-specified vmin, vmax arguments for phase map plot

neilzim:
Currently, when you display a phase map, user-specified vmin and vmax values are ignored. I added a few lines of logic so that when the user provides parameters that are not None, they override the defaults (-0.25, 0.25).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,7 @@ Code by Marshall Perrin, Joseph Long, Ewan Douglas, with additional
 contributions from Anand Sivaramakrishnan, Remi Soummer, Christine Slocum,
 and others on the Astropy team.
 
+test edit
 
 
 Projects using poppy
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -24,6 +24,7 @@ Code by Marshall Perrin, Joseph Long, Ewan Douglas, with additional
 contributions from Anand Sivaramakrishnan, Remi Soummer, Christine Slocum,
 and others on the Astropy team.
 
+test edit
```
josePhoenix: Could you remove this?

josePhoenix:
Can you write a sentence or so for the docstring saying what happens in the default case (0, max(data) for intensity; -.25, .25 for phase it looks like?)

josePhoenix:
Since the build with actual code changes passed, I'm going ahead and merging before Travis completes.

mperrin:
I didn't have a chance to look at this before you merged it. A small suggestion would be to add to the docstring for vmin,vmax another sentence as follows:  "If the scale parameter is set to 'log', then the default vmin for intensity becomes 1e-6*intens.max() instead of 0."

PR #184, opened 2016-09-08
added logarithmic case to the docstring about the vmin and vmax param…

neilzim:
…eters for the Wavefront display() method


mperrin:
Thanks!

PR #189, opened 2016-09-21
Change Travis to use ci-helpers

eteq:
This is based on an initial effort by @josePhoenix with some follow-up on my part. It updates the .travis.yml to use the newer `ci-helpers` infrastructure, which is generally easier to read because it's build around shell/python scripts instead of arcane `.travis.yml` machinery.

Note that this also removes the `setup_requires` from `setup.py` - that's necessary to get `egg_info` to run without trying to install numpy.  This is definitely important behavior in certain contexts, so I think it's desired, but it's *possible* that it has some unintended consequences if `poppy` is doing something unusual.  I suspect, though, that it's a throw-back to before the `astropy-helpers`, when all of its machinery was packaged as part of astropy itself.


eteq:
Oh, and in its current form this doesn't do coveralls... let me know if I should turn it on (I wasn't clear on whether `poppy` is actually using it or not)

josePhoenix:
POPPY was set up at one point to use Coveralls. It seems to have come back to life with a recent change to the test infrastructure, so we'd like it to work. May be out of scope.

My concern with this `.travis.yml` is that it works but still does the wrong thing (namely installing `enum34` in situations where it will clobber a standard library module). (To be fair, the current build does that too, which is part of why it works.)

I can see how I would hack Travis with some environment variables and conditionals to install more stuff, but I cannot see how to express such a conditional dependency with the Astropy CI helpers.

eteq:
> namely installing enum34 in situations where it will clobber a standard library module

Are you sure it clobbers the stdlib module?  I just tried installing it in py 3.5 and it appears the stdlib module takes precedence...

I could also modify this to only use enum in the 2.7 versions, though, if you like.  The problems we were seeing before I think were due to the `setup_requires` bits, not an enum dependency problem.

eteq:
Aha! the last commit there just turned coveralls back on, so that's working.  I'm speculating you want coveralls on py 3.5, but of course that's a trivial change to switch to 2.7.

Also, check out https://travis-ci.org/eteq/poppy/builds/161755807 for a version that only uses enum34 on 2.7 - I can switch this over to that if you like.

.travis.yml
```diff
@@ -1,113 +1,101 @@
...
     # The coveragerc file may be customized as needed for your package.
     - if [[ $SETUP_CMD == 'test --coverage' ]]; then coveralls --rcfile='poppy/tests/coveragerc'; fi
-
 notifications:
-    email:
-        - mperrin@stsci.edu
-        - jlong@stsci.edu
+  email:
+    # - mperrin@stsci.edu
+    - jlong@stsci.edu
```
mperrin: Leave me on the Travis notifications for this please.

mperrin:
Coveralls on 3.5 is correct, yeah. Going to deprecate 2.7 as soon as astropy does, so everything should be in 3 by default.

eteq:
ok, @mperrin is back on the list.  @josePhoenix, do you prefer the version that moves enum34 to only py27 (https://github.com/mperrin/poppy/compare/master...eteq:travis-ci-helpers-test) or this one?

josePhoenix:
I'm a fan of the one showing Python 2.7 customizations. Even if it's not strictly necessary, I suspect we'll grow more backported dependencies the further we get into Python-3-first development.

mperrin:
@eteq meanwhile I've started seeing a strange error in Travis on webbpsf related to astropy-helpers, see for instance https://travis-ci.org/mperrin/webbpsf/jobs/164074052

```
    /tmp/pip-build-YgWNZk/poppy/astropy_helpers/astropy_helpers/setup_helpers.py:100: AstropyDeprecationWarning: Direct use of the adjust_compiler function in setup.py is deprecated and can be removed from your setup.py.  This functionality is now incorporated directly into the build_ext command.
      'command.', AstropyDeprecationWarning)
```

Do you have any insights into what's causing that? It only happens on 2.7 but not 3.5, and it seems to have started recently for reasons that are unclear. Would this switch to ci-helpers fix that?

josePhoenix:
At risk of mixing up issues in different threads, I have a first whack at `.travis.yml` for WebbPSF with the helpers. Still needs some tweaks, but it's building on 2.7 now. https://travis-ci.org/josePhoenix/webbpsf/jobs/164089397

PR #195, opened 2016-10-04
Fix propagate_direct pixel scale units, as noted in #194

douglase:


mperrin:
This is a nice simple fix and I'll merge it once the test suite passes on Travis. But this does show that our test suite isn't complete enough for the Fresnel code... That's something for us to fix in future PRs I hope!

PR #203, opened 2017-02-18
Fix for non-integer indexing in Numpy >= 1.12.0

kmdouglass:
As of version 1.12.0, NumPy no longer implicitly casts non-integers to integers for array indexing and instead raises an `IndexError`. See the [Numpy Release Notes](https://docs.scipy.org/doc/numpy-dev/release.html#id6) for details.

A couple of lines in poppy_core.py and the test scripts required either explicit integer casting or integer division with the `//` operator for compatibility with NumPy >= 1.12.0.

The Poppy test suite is now passing on my machine with Python 3.5.2 and 2.7.13 and Numpy 1.12.0 with these fixes. You may want to add a build with Numpy 1.12.0 to Travis to verify that things work on your end.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2273,10 +2273,8 @@ class OpticalElement(object):
                 lx,ly=resampled_amplitude.shape
                 #crop down to match size of wavefront:
                 lx_w,ly_w = wave.amplitude.shape
-
-                # explicit cast to int fixes index errors in Numpy >= 1.12.0
-                border_x = int(np.abs(np.floor((lx-lx_w)/2)))
-                border_y = int(np.abs(np.floor((ly-ly_w)/2)))
+                border_x = np.abs(lx-lx_w) // 2
+                border_y = np.abs(ly-ly_w) // 2
                 if (self.pixelscale*self.amplitude.shape[0] < wave.pixelscale*wave.amplitude.shape[0]) or (self.pixelscale*self.amplitude.shape[1] < wave.pixelscale*wave.amplitude.shape[0]):
                     #raise ValueError("Optic is smaller than input wavefront")
                     _log.warn("Optic"+str(np.shape(resampled_opd))+" is smaller than input wavefront"+str([lx_w,ly_w])+", will attempt to zero-pad the rescaled array")
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2273,8 +2273,10 @@ class OpticalElement(object):
                 lx,ly=resampled_amplitude.shape
                 #crop down to match size of wavefront:
                 lx_w,ly_w = wave.amplitude.shape
-                border_x = np.abs(np.floor((lx-lx_w)/2))
-                border_y = np.abs(np.floor((ly-ly_w)/2))
+
+                # explicit cast to int fixes index errors in Numpy >= 1.12.0
+                border_x = int(np.abs(np.floor((lx-lx_w)/2)))
+                border_y = int(np.abs(np.floor((ly-ly_w)/2)))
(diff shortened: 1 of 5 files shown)


poppy/poppy_core.py
```diff
@@ -2273,8 +2273,10 @@ def get_phasor(self,wave):
                 lx,ly=resampled_amplitude.shape
                 #crop down to match size of wavefront:
                 lx_w,ly_w = wave.amplitude.shape
-                border_x = np.abs(np.floor((lx-lx_w)/2))
-                border_y = np.abs(np.floor((ly-ly_w)/2))
+
+                # explicit cast to int fixes index errors in Numpy >= 1.12.0
+                border_x = int(np.abs(np.floor((lx-lx_w)/2)))
```
mperrin: I'm on the fence about whether we need explanatory comments on each instance of this fix. It makes it clearer now what the logic is for the changes, but thinking about this from the perspective of code maintenance 2-5 years from now I'm not sure it enhances code readability to keep drawing attention to the index errors then. @josePhoenix, @douglase opinions on what's best for signal to noise here?

poppy/tests/test_nonsquare.py
```diff
@@ -55,9 +55,12 @@ def test_nonsquare_detector_values(oversample=1, pixelscale=0.010, wavelength=1e
 
         #pl.figure(2)
         psf0 = results[0]
+        print(psf0)
```
mperrin: Extra print statement is unnecessary :-)

mperrin:
Very nice, thank you @kmdouglass. I have to admit I hadn't even realized numpy 1.12 was out yet. Much appreciated!  I'll give this PR a day or so for code review & feedback from others here (not that it needs much) then pending that I expect this will be good to merge.

kmdouglass:
You're welcome @mperrin . If there's something else that you need me to do, just let me know.

poppy/poppy_core.py
josePhoenix: I'm indifferent about the comments.

However, for this line, there are three function calls where there should be a single operator:

```
In [12]: lx_w, lx = 20, 45

In [13]: int(np.abs(np.floor((lx-lx_w)/2)))
Out[13]: 12

In [14]: (lx - lx_w) // 2
Out[14]: 12
```

Since the `shape` elements are integers, integer division is the way to go.

poppy/poppy_core.py
```diff
@@ -2273,8 +2273,10 @@ def get_phasor(self,wave):
...
+                border_y = int(np.abs(np.floor((ly-ly_w)/2)))
```
josePhoenix: Same thing here.

poppy/tests/test_nonsquare.py
josePhoenix: :+1:

poppy/poppy_core.py
mperrin: Ah, but on the other hand: 
```
In [1]: lx_w, lx = 45, 20

In [2]: int(np.abs(np.floor((lx-lx_w)/2)))
Out[2]: 13

In [3]: (lx - lx_w) // 2
Out[3]: -13
```

We know the shape elements are integers but we don't know which is larger.  So I think the absolute value is necessary.

kmdouglass:
Thanks for the feedback. If I understood everyone correctly I will amend this pull request by

1. removing the comments (I agree with @mperrin that they are a bit too verbose. NumPy makes it very clear why there is an indexing error if these are ever changed in the future);
2. amend the `int(np.abs(np.floor((lx-lx_w)/2)))` call to `np.abs(lx-lx_w) // 2`, leaving the np.abs call because the order of `lx` and `lx_w` is not known a priori;
3. remove the extraneous `print` statement (I'm sorry about this; I'll be more careful in the future.)

If this all seems correct then I'll make the changes. Thanks guys!

mperrin:
That all sounds correct to me.

(And no need to apologize for an extra print statement - not a big deal. We all do this sort of thing; a big part of why Github is so great is that it makes this kind of code review easy and efficient. All code needs more than one pair of eyes on it!)

PR #205, opened 2017-02-23
Pixel scale uniformity and FITS keyword documentation

douglase:
addresses #202 by adding a documentation page, increases the number of combinations of keywords checked for in _find_pixelscale_in_headers() and saves a new "pixunit" keyword.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -290,10 +290,12 @@ class Wavefront(object):
             else:
                 outFITS[0].header['FOV_X'] = (fov_arcsec[1], 'Field of view in arcsec (full array), X direction')
                 outFITS[0].header['FOV_Y'] = (fov_arcsec[0], 'Field of view in arcsec (full array), Y direction')
+                outFITS[0].header['PIXUNIT'] = 'arcsecond'
+
         else:
             outFITS[0].header['PIXELSCL'] = (self.pixelscale.to(u.meter/u.pixel).value, 'Pixel scale in meters/pixel')
             outFITS[0].header['DIAM'] =  (self.diam.to(u.meter).value, 'Pupil diameter in meters (not incl padding)')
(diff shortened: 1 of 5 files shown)


douglase:
Is anything additional needed to merge this?

poppy/poppy_core.py
```diff
@@ -2738,10 +2740,10 @@ def _find_pixelscale_in_headers(keywords, headers):
             if pixelscale is None and self.planetype is None:
                 # we don't know which keywords might be present yet, so check for both keywords
                 # in both header objects (at least one must be non-None at this point!)
-                _log.debug("  Looking for 'PUPLSCAL' or 'PIXSCALE' in FITS headers to set "
+                _log.debug("  Looking for 'PUPLSCAL' or 'PIXSCALE' or 'PIXELSCL' in FITS headers to set "
                            "pixel scale")
                 keyword, self.pixelscale = _find_pixelscale_in_headers(
-                    ('PUPLSCAL', 'PIXSCALE'),
+                    ('PUPLSCAL', 'PIXSCALE','PIXELSCL'),
```
mperrin: I suggest leaving this part of the code unchanged. For back compatibility, we should still allow PIXELSCL in input files we are reading in, even if a different keyword is used in the output files now. We don't want to break the ability to read files which we can currently read ok.

mperrin: Wait never mind I read that backwards, you did this right and I confused before and after in the diff. 

Too many hours in a row moving mirrors at Goddard, I'm not thinking straight.

mperrin:
Mostly just getting my attention again long enough to merge is. Thanks @douglase !

poppy/poppy_core.py
douglase: phew, you had me second guessing

PR #214, opened 2017-03-26
ignore nan values in radial standard deviations

douglase:
This change ignores NaN values passed to radial_profile(), effectively ignoring (treating as zero) in the cumulative sum and ignoring in the calculation of the standard deviation.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -2368,7 +2368,6 @@ class OpticalElement(object):
                 units = "\n".join(textwrap.wrap(units,20))
 
         if self.pixelscale is not None:
-            # TODO handle units better here for pupil vs. image planes? meters/pix vs arcsec/pix
             if self.pixelscale.decompose().unit ==u.m/u.pix:
                 halfsize = self.pixelscale.to(u.m/u.pix).value * self.amplitude.shape[0] / 2
             elif self.pixelscale.decompose().unit == u.radian/u.pix:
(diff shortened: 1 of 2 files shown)


douglase:
adding a commit handling unit conversion for display, this and above are changes previously present in pull request #185 but useful generally.

poppy/poppy_core.py
```diff
@@ -2369,7 +2369,13 @@ def display(self, nrows=1, row=1, what='intensity', crosshairs=True, ax=None, co
 
         if self.pixelscale is not None:
             # TODO handle units better here for pupil vs. image planes? meters/pix vs arcsec/pix
-            halfsize = self.pixelscale.value * self.amplitude.shape[0] / 2
+            if self.pixelscale.decompose().unit ==u.m/u.pix:
```
mperrin: Nice! You can probably remove the "TODO" comment on line 2371, since you have now done it. Thanks!

mperrin:
This looks good, but I'm curious under what circumstances you're getting NaNs in something you want to calculate a radial profile of?  In particular treating it as a zero in the sum will make the code not just return NaN for the whole profile, but it's not clear to me how meaningful the resulting profile will be if it's got a bunch of missing pixels.   ?

douglase:
The NaN scenario is wanting the standard deviation ignoring masked (saturated) pixels in diffraction spikes.

poppy/poppy_core.py
douglase: removed.

mperrin:
Could you please add to the doc strings of the function some brief description of this functionality with NaNs? I'm happy to merge it but we should try to make sure these behaviors are documented for interested users. Thanks!

douglase:
@mperrin, Is commit above what you were looking for in the doc-string?

PR #215, opened 2017-03-29
Add display amplitude

maciekgroch:
Finally I had time to fix that.

### DIFF
--- a/poppy/poppy_core.py
+++ b/poppy/poppy_core.py
@@ -536,8 +536,29 @@ class Wavefront(object):
 
             plot_axes = [ax1, ax2]
             to_return = (ax1, ax2)
+        if what == 'amplitude':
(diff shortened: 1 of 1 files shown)


poppy/poppy_core.py
```diff
@@ -536,8 +536,29 @@ def wrap_lines_title(title):
...
+            utils.imshow_with_mouseover(
+                amp,
+                ax=ax,
+                extent=extent,
+                norm=norm_inten,
+                cmap=cmap_inten,
+                origin='lower'
+            )
+            if title is None:
+                title = wrap_lines_title("Intensity " + self.location)
```
mperrin: Looks like the title string here needs to be changed to "Amplitude"

mperrin:
Thanks @maciekgroch !

PR #222, opened 2017-05-17
Fix arbitrary_basis clipping aperture

kvangorkom:
Addresses #221.

The implementation here varies slightly from what I did over in #221. Rather than pad the aperture array, I'm calculating what the new array size should be, passing this to `zernike_basis` as `npix`, and then cutting the circular basis down before the aperture is cut out of it and the basis re-orthonormalized.

To handle the optional `rho` and `theta` arguments, I'm padding these arrays with nonsense values. I _think_ this shouldn't have any negative consequences, but I am a bit uncomfortable with handling them this way.

(Also adds a check that the aperture is passed in as a square array.)

### DIFF
--- a/poppy/tests/test_zernike.py
+++ b/poppy/tests/test_zernike.py
@@ -140,6 +140,64 @@ def test_cross_hexikes():
         _test_cross_hexikes(testj=testj, nterms=6)
 
 
+def test_arbitrary_basis_rms(nterms=10, size=500):
+    """Verify RMS(Zernike[n,m]) == 1."""
+    # choose a square aperture for the test
+    square_aperture = optics.SquareAperture(size=1.).sample(npix=size,grid_size=1.1)
+    square_basis = zernike.arbitrary_basis(square_aperture,nterms=nterms)
+
+    assert np.nanstd(square_basis[0]) == 0.0, "Mode(j=0) has nonzero RMS"
+    for j in range(1, nterms):
+        rms = np.nanstd(square_basis[j])  # exclude masked pixels
+        assert abs(1.0 - rms) < 0.001, "Mode(j={}) has RMS value of {}".format(j, rms)
+
+
+def _test_cross_arbitrary_basis(testj=4, nterms=10, npix=500):
+    """Verify the functions are orthogonal, by taking the
+    integrals of a given mode times N other ones.
+
+    Parameters :
+    --------------
+    testj : int
+        Index of the Zernike-like polynomial to test against the others
+    nterms : int
+        Test that polynomial against those from 1 to this N
+    npix : int
+        Size of array to use for this test
+    """
+    # choose a square aperture for the test
+    square_aperture = optics.SquareAperture(size=1.).sample(npix=npix,grid_size=1.1)
+    square_basis = zernike.arbitrary_basis(square_aperture,nterms=nterms)
+
+    test_mode = square_basis[testj - 1]
+    for idx, array in enumerate(square_basis):
+        j = idx + 1
+        if j == testj or j == 1:
+            continue  # discard piston term and self
+        prod = array * test_mode
+        wg = np.where(np.isfinite(prod))
+        cross_sum = np.abs(prod[wg].sum())
+
+        # Threshold was originally 1e-9, but we ended up getting 1.19e-9 on some machines (not always)
+        # this seems acceptable, so relaxing criteria slightly
+        assert cross_sum < 2e-9, (
+            "orthogonality failure, Sum[Mode(j={}) * Mode(j={})] = {} (> 2e-9)".format(
+                j, testj, cross_sum)
+        )
+
+
+def test_cross_arbitrary_basis():
+    """Verify orthogonality for a subset of Hexikes by taking the integral of
+    that Hexike times N other ones.
+
+    Note that the Hexike are only strictly orthonormal over a
+    fully hexagonal aperture evauated analytically. For any discrete
+    aperture the orthonormality is only approximate.
+    """
+    for testj in (2, 3, 4, 5, 6):
+        _test_cross_arbitrary_basis(testj=testj, nterms=6)
+
+
 def test_opd_expand(npix=512, input_coefficients=(0.1, 0.2, 0.3, 0.4, 0.5)):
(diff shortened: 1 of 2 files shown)


poppy/zernike.py
```diff
@@ -702,17 +702,42 @@ def arbitrary_basis(aperture, nterms=15, rho=None, theta=None):
...
+        "only square aperture arrays are supported"
+
+    # To avoid clipping the aperture, we precompute the zernike modes
+    # on an array oversized s.t. the zernike disk circumscribes the 
+    # entire aperture. We then slice the zernike array down to the 
+    # requested array size and cut the aperture out of it.
+
+    # get max extent of aperture from array center
+    yind, xind = np.where(aperture > 0)
+    distance = np.sqrt( (yind-shape[0]/2.)**2 + (xind-shape[1]/2.)**2 )
```
mperrin: This is somewhat subtle but you actually want `yind-(shape[0]-1)/2` here, and likewise for x. Consider::

``` python
In [6]: shape=(3,3)

In [7]: yind,xind=np.indices(shape)

In [8]: yind-shape[0]/2.
Out[8]:
array([[-1.5, -1.5, -1.5],
       [-0.5, -0.5, -0.5],
       [ 0.5,  0.5,  0.5]])

In [9]: yind-(shape[0]-1)/2.
Out[9]:
array([[-1., -1., -1.],
       [ 0.,  0.,  0.],
       [ 1.,  1.,  1.]])

```

poppy/zernike.py
```diff
@@ -702,17 +702,42 @@ def arbitrary_basis(aperture, nterms=15, rho=None, theta=None):
...
+    padding = ( int(ceil((max_extent - shape[0] / 2.))),
+                int(ceil((max_extent - shape[1] / 2.))) )
+    padded_shape = (shape[0] + padding[0] * 2, shape[1] + padding[1] * 2)
+    npix = padded_shape[0]
+
+    # pad theta and rho arrays
+    if theta is not None:
+        theta = np.pad(theta,padding,mode='constant',constant_values=0.)
+    if rho is not None:
+        rho = np.pad(rho,padding,mode='constant',constant_values=1.)
```
mperrin: I agree this handling of `theta` and `rho` doesn't feel right. As I recall the intent of this code is a very flexible capability for users to hand in arbitary indices to specific pixels without actually having a full contiguous array (consider a case of a sparse pupil which is mostly zeros; it would be more efficient to only evaluate zernike polynomials for the subset of pixels with nonzero intensity.)

Note that if `rho` and `theta` are not None, the `npix` argument is completed ignored inside `zernike_basis`.  Probably what I'd do here myself is check whether these arguments are present, and if so, just skip all the padding steps entirely. It's an advanced mode so it seems safe to declare that users who are using those arguments ought to be responsible for their own aperture padding.

mperrin:
Overall looks good. I made a couple comments in-line on the code. 

If you're feeling ambitious, would you be willing to put together a unit test for this function that demonstrates it works as desired? The coveralls check reminded me that we don't have any test code for `arbitrary_basis` at all yet...

kvangorkom:
The latest commits should hopefully address all your comments!

For unit tests, I just duplicated the tests that were being performed for the other bases--one check for the normalization and one for orthogonality.

A few things I noticed that probably deserve their own issues :

1.  I'm not sure that `test_zernikes_rms` is performing the check quite correctly. The line
```
assert 1.0 - rms < 0.001, "Zernike(j={}) has RMS value of {}".format(j, rms)
```
ought to be
```
assert abs(1.0 - rms) < 0.001, "Zernike(j={}) has RMS value of {}".format(j, rms)
```
correct? Otherwise, RMS values > 1 sneak on by. (The latter is how I formulated it in `test_arbitrary_basis_rms`)

2. I'm not sure the different combinations of `rho`, `theta` being `None` are handled correctly (`theta=None` with `rho` defined hits an array broadcasting error, `rho=None` with `theta` defined runs successfully with unexpected results). I get the same behaviors with `zernike_basis` as with `arbitrary_basis`, so I left it alone for now. 
3. `hexike_basis`  ignores the `outside` argument.

poppy/tests/test_zernike.py
```diff
@@ -140,6 +140,64 @@ def test_cross_hexikes():
...
+def test_cross_arbitrary_basis():
+    """Verify orthogonality for a subset of Hexikes by taking the integral of
+    that Hexike times N other ones.
+
+    Note that the Hexike are only strictly orthonormal over a
+    fully hexagonal aperture evauated analytically. For any discrete
+    aperture the orthonormality is only approximate.
+    """
+    for testj in (2, 3, 4, 5, 6):
+        _test_cross_arbitrary_basis(testj=testj, nterms=6)
```
mperrin: Minor: This is somewhat inefficient as written, because it's going to regenerate the basis from scratch inside each call to `_test_cross_arbitrary_basis`. Not sure how much this matters though (we're not paying for Travis CI CPU time...).

poppy/tests/test_zernike.py
```diff
@@ -140,6 +140,64 @@ def test_cross_hexikes():
...
+        # this seems acceptable, so relaxing criteria slightly
+        assert cross_sum < 2e-9, (
+            "orthogonality failure, Sum[Mode(j={}) * Mode(j={})] = {} (> 2e-9)".format(
+                j, testj, cross_sum)
+        )
+
+
+def test_cross_arbitrary_basis():
+    """Verify orthogonality for a subset of Hexikes by taking the integral of
+    that Hexike times N other ones.
```
mperrin: Very minor: comment text says "Hexike" but should just say "basis function"

mperrin:
Nice work. For your suggested other issues those all look like good catches, thanks very much. I'll spawn a new issue and fix those myself.

mperrin:
This all looks good, so I'm going to go ahead and merge. Thanks much @kvangorkom !

poppy/tests/test_zernike.py
mperrin: I see the inefficiency is my own, replicated in both the zernike and hexike tests!  And is not a big deal since the test case execution time (measured just now on my Mac Pro) is just ~ 0.5s for zernike, 0.6s for hexike, and ~ 1s for arbitrary basis.

PR #224, opened 2017-05-23
Fix misc issues from #223

mperrin:
@kvangorkom Take a look and see if you think this addresses the points you raised. 

(Ironically, while adding the `outside` option to the hexike function, at one point I had a bug such that `hexike_basis` always returned all NaN for all pixels always. And the whole test suite passed anyway! Because all the comparisons were only checking the valid non-Nan pixels. Ha.  So I added a new test case for that too. )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 
 python:
     - 2.7
-    - 3.5
+    - 3.6
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
 sudo: false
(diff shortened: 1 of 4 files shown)


kvangorkom:
Hm. `poppy.zernike.zernike_basis(theta=theta,rho=None)` is still sneaking by.  It looks like `zernike_basis` checks only for `rho`, and, if it doesn't see it, silently tosses out the `theta` argument and circumvents the check you added in `zernike`. (So it essentially behaves as if you didn't pass in either one.)

The `outside` option doesn't appear to be quite right—the Gram-Schmidt process clobbers any value adopted before the re-orthonormalization. In `arbitrary_basis`, I simply added an extra line at the end to do this (as well as cast the list of terms to an ndarray for consistency with `zernike_basis`).

mperrin:
After much delay I went back and implemented the fixes you suggested here, @kvangorkom. Thanks for the helpful close reading of my code!  Shouldn't have taken this long to fix, but I hadn't had time to work on webbpsf or poppy at all for the last two months until today (when it became critical suddenly since I had to fix breakage from the astropy 2.0 release).

PR #234, opened 2017-08-18
Requested feature - added option to return complex image plane wavefront 

douglase:
suggested by @jjlumbres, return the final complex plane with calc_psf, without returning all the other intermediate wavefront planes, saving memory.

w/ tests.


poppy/tests/test_core.py
```diff
@@ -393,3 +393,11 @@ def test_unit_conversions():
...
         assert np.all(np.abs(difference) < 3e-5)
 
+def test_return_final_wf():
+    osys =poppy_core.OpticalSystem()
+    osys.add_pupil(optics.CircularAperture(radius=3))   
+    osys.add_detector(pixelscale=0.010, fov_arcsec=5.0)
+    psf = osys.calc_psf(2e-6,return_final=True) 
+    assert len(psf[1])==1 #make sure only one element was returned
+    #test that the wavefront returned is the final wavefront:
+    assert np.max(psf[1][0].intensity-psf[0][0].data)**2 ==0
```
mperrin: Another way of doing this test that might be a little better would be
```assert np.allclose(psf[1][0].intensity, psf[0][0].data)```

Using `allclose` is more robust since it would ignore any differences around machine precision in the least significant bits. I agree it seems reasonable to expect exact agreement here but I'm not 100% certain that would hold try in all cases. 

I'm also a little confused about the `**2` since that's outside the call to `np.max()`; is that intentional?

File: poppy/tests/test_fresnel.py [hunk 2]
```diff
@@ -446,3 +446,21 @@ def test_fresnel_propagate_direct_2forward_and_back():
...
+    gl=fresnel.QuadraticLens(500*u.cm)
+
+    tel.add_optic(gl)
+    tel.add_optic(optics.CircularAperture(radius=M1_radius,name="M1 aperture"))
+    tel.add_optic(optics.ScalarTransmission( name="primary mirror focal plane"), distance=fl_M1)
+    
+    psf=tel.calcPSF(return_final=True)
+    
+    assert len(psf[1])==1
+    assert np.max(psf[1][0].intensity-psf[0][0].data)**2 ==0
```
mperrin: same comment as for `test_return_final_wf`.

poppy/tests/test_fresnel.py
```diff
@@ -446,3 +446,21 @@ def test_fresnel_propagate_direct_2forward_and_back():
     wf.propagate_direct(z)
     wf.propagate_direct(-z)
     np.testing.assert_almost_equal(wf.wavefront, start)
+
+def test_fresnel_return_complex():
```
mperrin: consider renaming these two test functions so the names are more parallel, to make it obvious that it's the same basic test for Fraunhofer and Fresnel. `test_return_final_complex_wavefront` and  `test_fresnel_return_final_complex_wavefront`?

mperrin:
This looks good, and I can see why this feature is a good idea, sure. 

Semi-relatedly, this PR made me realize about an API inconsistency: we use `retain_intermediates` in `poppy_core.py` but `return_intermediates` in `instrument.py`. I wonder if it's worth doing something to make that more consistent. But this almost certainly isn't the time to do that right now in any case.

poppy/tests/test_core.py
douglase: **2 was supposed to be inside the max. good catch.

I switched the assert to allclose, though since it was working on the same array I really think it shoudn't change.

douglase:
(@mperrin, the square was because I used max, I didn't want an array with a bunch of negative numbers and one zero to pass the test.)

douglase:
Is further revision required to merge this PR?

mperrin:
Eh, mostly just your reminding me. Thanks :-) 

I sort of vaguely thought there was some check that wasn't passing, but that doesn't appear to be the case. I must have confused it with a different PR I guess.  Happy to merge it now.

mperrin:
Huh. Despite both tests passing, merging this made the Travis build of `master` fail, in the step where it builds the docs. Huh? Going to try to kick it and see if re-running that job does any better

PR #237, opened 2017-10-03
fix documentation of divergence angle

douglase:
the returned divergence is the full angle not the half angle as previously implied. See definition in Kogelnik and Li, 1966, eq. 22, or the Newport Gaussian Beam Technical note (https://www.newport.com/n/gaussian-beam-optics)

Kogelnik, H., and T. Li. 1966. “Laser Beams and Resonators.” Applied Optics 5 (10): 1550–67. doi:10.1364/AO.5.001550.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -307,11 +307,11 @@ class FresnelWavefront(Wavefront):
     @property
     def divergence(self):
         """
-        Full-angle divergence of the gaussian beam
+        Half-angle divergence of the gaussian beam
 
-        I.e. twice the angle between the optical axis and the beam radius (at a large distance  from the waist) in radians.
+        I.e.  the angle between the optical axis and the beam radius (at a large distance  from the waist) in radians.
         """
-        return 2 * self.wavelength / (np.pi * self.w_0)
+        return self.wavelength / (np.pi * self.w_0)
 
     @property
     def param_str(self):
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -307,10 +307,9 @@ class FresnelWavefront(Wavefront):
     @property
     def divergence(self):
         """
-        Divergence of the gaussian beam
+        Full-angle divergence of the gaussian beam
 
-        I.e. the angle between the optical axis and the beam radius at a large distance.
-        Angle in radians.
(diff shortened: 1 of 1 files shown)


mperrin:
Huh, good catch. The other possible fix of course would be to remove the '2' from that line, so that it does return the half angle instead of full angle. You think it's better to just change the docs? Looking through a couple different references just now it doesn't seem like people are very consistent about whether they're talking about full angle or half angle...  I'm fine with just fixing the docs to be consistent but wanted to at least ask the question.

douglase:
Kogelnik and Li drop the 2, as you suggest, and Newport and Verdeyen (Laser Electronics) define it as it is currently.  It doesn't look like it's called anywhere else in POPPY, I was hesitant to break anyone's code in case they were using it appropriately but I would be inclined to drop the 2 if you don't think anyone else is using it (correctly).

mperrin:
> It doesn't look like it's called anywhere else in POPPY, I was hesitant to break anyone's code in case they were using it appropriately but I would be inclined to drop the 2 if you don't think anyone else is using it (correctly).

In that case let me ping both @maciekgroch and @corcoted, who I know are using the Fresnel code. If either of you have got a spare moment would you be so kind as to take a look at this issue, and weigh in on which way you would rather see this inconsistency fixed?  Thanks in advance!

corcoted:
This factor of two is inconsistent throughout the laser literature, too.  As noted above, manufacturers tend to quote to full divergence angle and scientists tend to use the half angle.  For example, the half angle is easier to use when doing paraxial propagation using the complex beam parameter equations, but the full angle is the number that the beam profiler reports.

I prefer the half angle myself, but I think either is acceptable if we're consistent and clearly document it.

mperrin:
@douglase in your last comment on this you said "I would be inclined to drop the 2 if you don't think anyone else is using it (correctly).", and I think we don't have any objections to doing it that way. Would you like to go ahead and modify the PR to fix is that way instead?  Either way's fine with me, I just don't want to leave this an orphan hanging PR too long (like I did to your other one!)

douglase:
switched to half angle in 1c1ffca.

mperrin:
Thanks! I'm going to go ahead and merge this because all the actual code tests pass, apart from the known issue of the docs not building on Travis any more.

PR #239, opened 2017-11-13
CUDA accelerated propagation

douglase:
This is a proof of concept addition of to see if CUDA FFTs would significantly accelerate the angular spectrum code. While that is the case, in the course of implementation and benchmarking (profiling with ipython %prun) I discovered significant bottlenecks were np.fft.fftshift() and np.exp().  These are also addressed in this branch, the former with a numba gpu implementation fftshift and the latter using the NumExpr library. 

This obviously adds a lot of optional dependencies, but my philosophy was follow the fftw approach and gracefully revert to numpy everywhere, so each addition gives partial improvement. On my machine tests are passing with and w/o numba or cuda installed.

While further optimization is possible, for a toy case with 3 optical surfaces (below) generating a PSF with a fresnel wavefront class is accelerated by as much as a factor of 3:

image

Benchmark Optical system:
```

wavelen = 770e-9

D_prim = 2.37 * u.m
fr_pri = 7.8
fl_pri = D_prim * fr_pri

D_relay = 20 * u.mm
fl_m2 = fl_pri * D_relay / D_prim
fr_m3 = 20.
fl_m3 = fr_m3 * D_relay

def wfirst_sys(npix=1024,ratio=0.25):
    wfirst_optsys = poppy.FresnelOpticalSystem(pupil_diameter=D_prim, npix=npix, beam_ratio=ratio)

    m1 = poppy.QuadraticLens(fl_pri, name='Primary')
    m2 = poppy.QuadraticLens(fl_m2, name='M2')
    m3 = poppy.QuadraticLens(fl_m3, name='M3')
    m4 = poppy.QuadraticLens(fl_m3, name='M4')

    wfirst_optsys.add_optic(poppy.CircularAperture(radius=D_prim.value/2))
    wfirst_optsys.add_optic(m1)
    wfirst_optsys.add_optic(m2, distance = fl_pri + fl_m2)
    wfirst_optsys.add_optic(m3, distance = fl_m2 + fl_m3)

    #wfirst_optsys.add_optic(poppy.ScalarTransmission(planetype=poppy.poppy_core.PlaneType.image, name='focus'),
    #                        distance=fl_m3)

    wfirst_optsys.add_optic(poppy.ScalarTransmission(planetype=poppy.poppy_core.PlaneType.intermediate,
                                                 name='focus'), distance=fl_m3)
    return wfirst_optsys
wfirst_optsys4096=wfirst_sys(npix=1024,ratio=0.25)
%timeit wfirst_optsys4096.calcPSF(wavelength=wavelen, display_intermediates=False, return_intermediates=False)
```

Benchmark computer:

Google Compute Engine:

> 
> Machine type
> 
> n1-standard-8 (8 vCPUs, 30 GB memory)
> CPU platform
> 
> Intel Haswell
> GPUs
> 
> 1 x NVIDIA Tesla K80
>


douglase:
also, h/t to @neilzim for the benchmark prescription.

I will investigate the travis failure to build.

mperrin:
This is spectacular! I've been wanting to dig into all these performance optimization toolkits but just haven't had time. In addition to cuda and numexpr did you do any tests with numba?   As I recall, @josePhoenix had done some tests with Cython but wasn't able to get performance better than using a well-optimized numpy for the array operations.  That was in the Fraunhofer code rather than the Fresnel though.   I'm excited to see you got a 3x speedup using cuda!

poppy/fresnel.py
```diff
@@ -256,6 +276,11 @@ def __init__(self,
         if self.oversample < 2:
             _log.warn("Oversampling > 2x suggested for reliable results.")
 
+        self._y, self._x = np.indices(self.shape, dtype=float)
+        self._y -= (self.wavefront.shape[0]) / 2.0
+        self._x -= (self.wavefront.shape[1]) / 2.0
+        """saves x and y indices for future use"""
```
mperrin: This is interesting. I'm guessing you decided to save these because benchmarking showed a significant enough cost in all the calls to generate the coordinates?  I'm a little surprised if there's substantial savings from doing this since you still have to do the multiplications by the pixel scales in the coordinates functions. But I have to admit I didn't actually bench mark it myself

poppy/fresnel.py
```diff
@@ -636,14 +676,20 @@ def _propagate_ptp(self, dz):
             return
 
         x, y = self.coordinates()  # meters
-        rhosqr = np.fft.fftshift((x / (self.pixelscale.to(u.m / u.pix).value ** 2 * self.n)) ** 2 + (
-                                  y / (self.pixelscale.to(u.m / u.pix).value ** 2 * self.n)) ** 2)
-        t = -1.0j * np.pi * self.wavelength.to(u.meter).value * (
-            z_direct) * rhosqr  # Transfer Function of diffraction propagation eq. 22, eq. 87
+        meter_per_pix = self.pixelscale.to(u.m / u.pix).value
+        rhosqr = np.fft.fftshift((x / ( meter_per_pix** 2 * self.n)) ** 2 + (
```
mperrin: Should this use your `_fftshift` function here?

poppy/tests/test_fresnel.py
```diff
@@ -196,7 +196,13 @@ def test_Circular_Aperture_PTP_short(display=False, npix=512, display_proper=Fal
...
         oversample=4)
     wf *= optics.CircularAperture(radius=800 * 1e-9*u.m)
-    wf_2 = wf.copy()
+    wf_2 = fresnel.FresnelWavefront(
+        2 * u.um,
+        wavelength=10e-9*u.m,
+        npix=npix,
+        oversample=4)
+    wf_2 *= optics.CircularAperture(radius=800 * 1e-9*u.m)
+
```
mperrin: Out of curiosity was there a reason you needed to duplicate this test code instead of copying the wavefront?

poppy/tests/test_fresnel.py
```diff
@@ -293,56 +299,57 @@ def test_fresnel_optical_system_Hubble(display=False):
...
-        assert wavefront.angular_coordinates == angular_coordinates, assert_message.format(
-            idx, "angular_coordinates", angular_coordinates, wavefront.angular_coordinates
-        )
-
-    spherical_flags = [False, True, True, False]
-    for idx, (wavefront, spherical) in enumerate(zip(waves, spherical_flags)):
-        assert wavefront.spherical == spherical, assert_message.format(
-            idx, "spherical", spherical, wavefront.spherical
-        )
+    if len(waves)>1:
```
mperrin: This `if` statement's a little cryptic. Do I infer correctly that you had to add it because the CUDA version otherwise would fail this test since it can't return intermediates?

poppy/tests/test_fresnel.py
douglase: wavefront doesn't deepcopy with the cuFFTPLAN plan, it can't be pickled. the plan isn't initialized unless the cuda flag is enabled but I wanted to pass the test with cuda enabled.

poppy/tests/test_fresnel.py
douglase: That's right, can added a comment to that effect.

poppy/fresnel.py
douglase: yep.

joseph-long:
Awesome stuff, @douglase! I remember last time I looked at this I was trying to target OpenCL, which alas seems to be dead in the water, and numba's CUDA support was still proprietary/paid-license only. Exciting to see that support has matured so much!

douglase:
(much of the benchmarking for this project is recreated in this notebookhttps://gist.github.com/douglase/3846e8f105cd9baec96706681d0b8ee5

mperrin:
Wow, those benchmarks are really striking. Numexpr is way out in front, and for much less programming effort than the more manual numba vectorize and cuda versions. 

I've tried a bit here in benchmarking and adding numexpr calls to parts of the Fraunhofer pathway too, and gotten some nice speedups there. But the real heavy lifting is still the calls to the numpy dot product, which doesn't fit easily into what numexpr can help with.

mperrin:
@douglase I'm assuming you're around the AAS this week? I'm going to try re-starting the build of this PR on Travis now to see if it works (it was never clear why it died before). The actual code worked fine in my local tests. If necessary maybe we can meet up for half an hour sometime this week to try to figure out how to get this running in the Travis CI setup?

mperrin:
I ended up just merging this from the command line manually - trying to do anything clever with pushing the conflict fix onto your branch first just led me down into an unproductive morass of inscrutable git error messages. Was much simpler to just manually merge into master instead.

PR #244, opened 2018-03-26
Added inverse of measure_ee function 

shanosborne:
Added function in poppy/utils.py and named it measure_radius_at_ee


mperrin:
Hi @shanosborne, this code looks good.  Before merging in a new utility function we should also: 
 * add a unit test that verifies proper functionality of the new function. This would be in `poppy/tests/test_utils.py`
 * Consider adding some mention of this new function somewhere in the documentation, or adding example code to one of the example tutorial notebooks. 

Those should all be pretty quick additions. In fact, probably very similar code can be used as an example of how to use it and also as the unit test. We can chat more when we meet later this week.

poppy/utils.py
```diff
@@ -683,6 +683,46 @@ def measure_ee(HDUlist_or_filename=None, ext=0, center=None, binsize=None):
...
+    radius: function
+        A function which will return the radius of a desired encircled energy.
+
+    Examples
+    --------
+    EE = measure_radius_at_ee("someimage.fits")
+    print "The EE is 50% at {} arcsec".format(EE(0.5))
+    """
+
+    rr, radialprofile2, EE = webbpsf.radial_profile(HDUlist_or_filename, ext, EE=True, center=center, binsize=binsize)
```
mperrin: Hi @shanosborne, I think on this line you don't need or want to have the `webbpsf.` before the function call, because it's calling a function in this same file `utils.py`.   The unit test is not passing on Travis (see for instance: https://travis-ci.org/mperrin/poppy/jobs/365727304) and it looks like this is why.

mperrin:
Hmm, interesting. Your fix fixed the tests in almost all build cases. But it's still failing on the oldest configuration we test (numpy 1.8, astropy 1.3), with a linear algebra error about a singular matrix when fitting the interpolation function. See https://travis-ci.org/mperrin/poppy/jobs/365794975

Now, those are fairly ancient versions of those packages (current is numpy 1.14, and astropy 3.0). So one way to approach this would be simply to increment our minimal required versions of those to something more recent.  The test passes fine on numpy 1.11 and 1.12.   Looks like they made some improvements in the linear algebra code to make it more robust sometime between those two versions.  This is not something I want to have to dig into deeply. But uncovering this sort of compatibility issue is precisely why we run the test suite on a whole bunch of different versions of the core packages.

PR #248, opened 2018-05-03
Basically a FresnelWavefront wrapper class that uses different unit convention

DaPhil:
Based on [this issue](https://github.com/mperrin/poppy/issues/225) I wrote some time ago I wrote a wrapper class for the FresnelWavefront class to use physical units for the wavefront (i.e. V/m) and intensity (i.e. W/m^2). I also added some additional capabilities like computing the beam radius, center, and quality factor based on 2nd moments (i.e. ISO norm conform). If the pull request is accepted, I can provide an OpticalElement class that represents a thermal blooming phase screen (for which physical units of the intensity is necessary since it is a non-linear effect). In its current state, the class works but is probably not conform to how the main authors want a class to work in poppy. For example, I did not make use of the astropy.units package whenever I could. My main intention is rather to get an impression if this class and a thermal blooming phase screen is interesting enough to be added to poppy. If so, I am happy to work on the class to conform to any programming guidelines, naming conventions etc.

Unrelated note: I have also expertise in Kolmogorov phase screens and have noticed that there is a branch dedicated to that where I can contribute if required.


mperrin:
Thanks! This looks very good. I am definitely happy to have this added in. 

As you've probably inferred this is a very different use case from our needs here that motivated me to start poppy (thermal blooming from high energy lasers is not a relevant concern when observing ultra-faint galaxies with a space telescope! :-) ) but I am very happy to have this package grow to address other use cases in physics too. It's very nice to have someone contributing expertise in an area of physics I'm less familiar with - thanks. 

I'll take a look through the code and offer any suggestions that come to mind.  The main thing I would want to see before merging this is some unit tests that verify the basic functionality works as desired. Probably can be adapted from the example cases in your demo notebook. But you'll be the best person to write those up.  Doesn't need to be 100% coverage (we don't have anywhere near that already), but I'd like at least one or two basic tests for the PhysicalFresnelWavefront class. If you're not familiar with writing tests for py.test let me know and I can assist, but the files in the tests subdirectory should I hope be clear enough examples. 

Units is a messy topic. There's about a dozen different frameworks for handling them in Python (see list at https://pint.readthedocs.io/en/latest/faq.html). We use the astropy one since that's the most common among astronomers, but I realize it's a new thing to learn for non-astronomers so can be a barrier to more general use. I don't mind your not using them in all possible places. Personally I like having units on the parameters rather than having to remember what units are on a bare float, but already the code is hit-or-miss on that for historical reasons rather than entirely consistent.

poppy/__init__.py
```diff
@@ -93,13 +93,15 @@ class Conf(_config.ConfigNamespace):
 from . import optics
 from . import misc
 from . import fresnel
+from . import wavefront
```
mperrin: let's consider renaming this `physical_wavefront.py` or similar. Plain `wavefront.py` makes it sound like it contains the definitions of the `Wavefront` class itself which it doesn't.

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+    n0 : float
+        Ambient refractive index.
+    """
+    
+    def __init__(self,
+                 beam_radius,
+                 units=u.m,
+                 rayleigh_factor=2.0,
+                 oversample=2,
+                 wavelength=1064.0e-9,
```
mperrin: Generally in the other functions I've tried to leave the defaults at round numbers rather than baking in particular science/engineering assumptions about wavelengths.  So I'm a hesitant about having this default to Nd:YAG.  I also don't like having the different Wavefront classes have different defaults for a common parameter; that's not the least surprising interface. But hmm, I just realized the defaults on the other wavefronts are 2 microns, which relates back to the main science wavelengths for NASA's JWST mission, so that's not a great default either.   Now I think I might want to make the default 1 micron for all 3 of the Wavefront, FresnelWavefront, and PhysicalFresnelWavefront classes, so they would be consistent.

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+                 npix=1024,
+                 M2=1.0,
+                 n0=1.00027398, # refractive index of air @ 15°C, lambda=1064nm
```
mperrin: Oh interesting. My own biases are showing again because all of our use cases are in vacuum. Hmm, now I'm wondering if there are other places in this package where we might want to consider air vs vacuum as a parameter.

mperrin: At the very least it probably makes sense to have the documentation string state that PhysicalFresnelWavefront expects a vacuum wavelength, but then scales it and internally does the calculation with the wavelength in air.

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+                oversample=oversample,
+                wavelength=wavelength*M2/n0, # scale the wavelength and apply refractive index
+                npix=npix,
+                **kwargs
+                )
+        self.lam = wavelength/n0 # its unscaled value is needed to compute M2
+        self.n0 = n0
+        self.npix = npix*oversample
+        
+        return
```
mperrin: This and other bare return statements (not returning any value) aren't necessary.

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+        return const.c * self.n0 * const.epsilon_0 * np.abs(self.amplitude)**2 / 2.0
+    
+    @property
+    def power(self):
+        """Power of the wavefront (W)."""
+        
+        intensity = self.intensity
+        dx = self.dx
+        
+        return dx**2*np.sum(intensity)
```
mperrin: This function could be simplified to just
```
return (self.dx**2) * self.total_intensity
```

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+    
+    def scale(self, P):
```
mperrin: `scale` is a very generic name for this function and doesn't make clear what is being scaled. Perhaps `scale_power` or `set_power`? 

Relatedly there's the `normalize` function in the regular Wavefront class. That's not relevant for a `PhysicalFresnelWavefront` so you could consider adding code to make that explicit, and ignore any attempt to do that kind of normalization on this class:  
```
   def normalize(self):
        # for PhysicalFresnelWavefronts use scale_power instead of 
        # normalizing total intensity to 1
        pass
```

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+        P : float
+            The desired power of the wavefront.
+        """
+        
+        P0 = self.power
+        self.wavefront *= np.sqrt(P/P0)
+        
+        return
+    
+    def propagate(self, z, attenuation_coeff=0.0, **kwargs):
```
mperrin: Let's try to keep the API consistent between this and the parent class. Instead of adding a new function named `propagate`, this should just be named `propagate_fresnel`. Then it would call the parent class function of the same name via `super()`:
```
        P = self.power
        super(PhysicalFresnelWavefront,self).propagate_fresnel(z, **kwargs)
        self.scale_power(P*np.exp(-attenuation_coeff*z.to(u.m).value))
```

Doing it that way would ensure that all propagations of the `PhysicalFresnelWavefront` would use the power-conserving propagation, instead of letting someone accidentally call the non-power-conserving one.

mperrin:
OK that's it for now. Very nice. Oh, and I see there is a conflict on `__init__.py`, presumably due to some of my own recent additions to that. So you will need to merge the latest master into this branch and resolve that conflict before I can merge this PR. 

Danke schön!

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+            Attenuation coefficient (m^-1).
+        """
+        
+        P = self.power
+        self.propagate_fresnel(z, **kwargs)
+        self.scale(P*np.exp(-attenuation_coeff*z.to(u.m).value))
+        
+        return
+    
+    def plot(self, idx=100):
```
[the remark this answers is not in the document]
DaPhil: I can't remember why I added it anyway. Probably to have other colors etc as I wanted it. So I will just remove it.

poppy/wavefront.py
```diff
@@ -0,0 +1,359 @@
...
+        for idx in range(30):
+            w_buf = w
+            
+            center_x, center_y = self.center(mask)
+            
+            mask[:, :] = 0.0
+            for idx_x in range(num):
+                for idx_y in range(num):
+                    if abs(x[idx_x] - center_x) < 3.0*w_x and abs(x[idx_y] - center_y) < 3.0*w_y:
+                        mask[idx_x, idx_y] = 1.0
```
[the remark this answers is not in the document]
DaPhil: Well, I crafted the code all by myself, I am not aware of any sample code for this computation. I should note that I just started to use Python (a year or so ago) so I am not quite familiar with all conventions and capabilities of the relevant packages. I was just happy that the code works (even if it is slow).

DaPhil:
I must note that this is my very first pull request ever. And I mean EVER. I am not very familiar with version control (nor am I a Python Pro). And yes, I am not familiar with writing unit tests. But I will have a look at the examples you mentioned and try to work it out.

mperrin:
Well in that case congratulations!  Learning enough of git and github to make a pull request is a significant accomplishment. :-)

I wrote up some notes about how to write a test on an earlier pull request:  
https://github.com/mperrin/poppy/pull/227#issuecomment-313123274
So you can also take a look at that.  And don’t hesitate to send any questions my way; I’ll be happy to help out.

mperrin:
Thanks! Yes I had seen your test cases commits come in earlier. Those look very good - if you hadn't told me before I would not guess this was your first time working with unit tests or pull requests! 

The CI system intentionally runs the same tests on a variety of versions of Python and the major libraries. The test is only failing on the setup with the very oldest versions.  From the log messages at https://travis-ci.org/mperrin/poppy/jobs/376381937, it looks like something in the older setup is upset to encounter a Unicode degree symbol in the comment string on line 40 of physical_wavefront.py.  I expect if you change `15°C` to `15 deg C` there, then this error will go away.   

(We could also try to diagnose more precisely why a Unicode character is incompatibly with the older versions, but honestly it's just a comment string so I think it's easier to just avoid this rather than digging in any deeper.)

DaPhil:
Well, only took me a week to get it right :)

PR #249, opened 2018-05-15
Kolmogorov turbulent phase screens

DaPhil:
Hi. I implemented and tested turbulent wavefront errors within Kolmogorov theory. Added a notebook that describes the basics. Hope this is of use to anyone.


mperrin:
A pleasant surprise to see this PR!  As you saw, I had started some code for this but never had time or priority to get it working properly. This contribution is definitely welcomed. 

I see that you made it so the WFE is recalculated automatically with a different random instance for each call to get_opd. That's one thing I had thought about doing differently, to support polychromatic calculations in which one would want different wavelengths to all encounter the same phase screens. Perhaps there could be an option to freeze the OPD to a fixed example of the Kolmogorov phase screen to support cases like that. I'll look through the code to see if I have any other comments or suggestions. But I wanted to ask what you thought of this idea.


Two of the CI system tests cases failed. One is just the same issue with a file encoding as before. Easy fix now, same as before.  The other was a failure to build the documentation; this confuses me since it seems totally unrelated to the code changes. I don't know what to make of that one.

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        edited by P. B. Ulrich and L. E. Wilson
+        (International Society for Optics and Photonics, 1990), pp. 183–192.
+        
+        G. Gbur, J. Opt. Soc. Am. A 31, 2038 (2014).
+        
+        D. L. Knepp, Proc. IEEE 71, 722 (1983).
+        """
+        
+        coordinates = wave.coordinates()
+        npix = coordinates[0].shape[0]
```
mperrin: this could be equivalently done with just `npix = wave.shape[0]`

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        sign : int
+            Sign of mirror symmetry. Must be either +1 or -1.
+        
+        References
+        -------------------
+        Eq. (65) in J. A. Fleck Jr, J. R. Morris, and M. D. Feit,
+        Appl. Phys. 10, 129 (1976).
+        """
+        
+        sign = float(sign)
```
mperrin, line 504: could add an error check:   
```
if ((sign != 1) and (sign ! = -1)):
    raise ValueError("sign must be either +1 or -1")
```
Perhaps unnecessary but it's an easy check

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        
+        References
+        -------------------
+        G. Gbur, J. Opt. Soc. Am. A 31, 2038 (2014).
+        
+        R. Frehlich, Appl. Opt. 39, 393 (2000).
+        """
+        
+        if not any(kind==item for item in ['Kolmogorov', 'Tatarski', 'van Karman', 'Hill']):
+            raise AttributeError('Kind of power spectrum not correctly defined.')
```
mperrin: `AttributeError` is intended for cases when an object attribute reference fails (e.g. `my_function(some_var.wrong_attribute_name)` -> `object 'some_var' has no attribute 'wrong_attribute_name'`.)

For problems with input parameters like this, more typically the exception to raise would be `ValueError`.

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        
+        q2 = (qx**2 + qy**2)/u.pixel**2
+        if kind=='van Karman':
+            if L0 is not None:
+                q2 += 1.0/L0.to(u.m)**2
+            else:
+                raise AttributeError('If van Karman kind of turbulent phase \
+                                     screen is chosen, the outer scale L_0 \
+                                     must be provided.')
+        q2[0, 0] = np.inf # this is to avoid a possible error message in the next line
```
mperrin, line 582: I don't follow this part; why is it helpful to set `q2[0,0]` to infinity? What error message does it prevent?

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        # create real-valued random numbers with required symmetry
+        a = self.rand_symmetrized(npix, 1)
+        b = self.rand_symmetrized(npix, -1)
+        
+        # create complex-valued random number with required variance
+        c = (a + 1j*b)/np.sqrt(2.0)
+        
+        return c
+    
+    def power_spectrum(self, wave, L0=None, l0=None, kind='Kolmogorov'):
```
mperrin: I worry a little about having two parameters named so similarly as `L0` and `l0`. Seems to make it easy to accidentally have typos or other minor mistakes that cause unintended problems. Would it be better to have more explicit names `outer_scale` and `inner_scale`?

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        """ Returns the spatial power spectrum.
+        
+        Parameters
+        -----------------
+        wave : wavefront object
+            Wavefront to calculate the power spectrum for.
+        
+        L0 : float
+            The outer scale of the turbulent eddies (m).
```
mperrin: Add comment. "The outer scale only affects the calculation results if kind='von Karman'". 

Also, I believe 'van Karman' should be 'von Karman' throughout, spelled with an o.

poppy/wfe.py
```diff
@@ -299,3 +300,343 @@ def get_opd(self, wave, units='meters'):
...
+        
+        l0 : float
+            The inner scale of the turbulent eddies (m).
```
mperrin: Add comment "The inner scale affects the calculation results if kind = 'von Karman', 'Tatarski', or 'Hill'"

poppy/tests/test_wfe.py
```diff
@@ -86,3 +89,68 @@ def test_ParameterizedAberration():
...
+    
+    assert(np.round(np.abs(np.mean(a)), 2) == np.round(0.0, 2))
+    assert(np.round(np.var(a), 2) == np.round(1.0, 2))
+    assert(np.round(np.mean(b), 2) == np.round(0.0, 2))
+    assert(np.round(np.var(b), 2) == np.round(1.0, 2))
+    assert(np.round(np.mean(c), 2) == np.round(0.0, 2))
+    assert(np.round(np.var(c), 2) == np.round(1.0, 2))
+
+def test_KolmogorovWFE_Cn2():
+    # verifiy correct calculation of Cn2 from Fried parameter
```
mperrin: very minor: wrong spelling of 'verify' in comment, here and in the following function

poppy/wfe.py
DaPhil: q2[0, 0] is zero (except for van Karman). The next line tries to divide by q2, so there would be a "division by zero" error. The value for q2[0, 0] is irrelevant really, because the random array at this position also has been set to 0.

DaPhil:
Based on [this issue](https://github.com/astropy/astropy-helpers/issues/302) but without any real knowledge of what I am speaking of, it seems that the astropy helpers need to be updated. However, I committed a changed astropy_helpers which had a changed _Subproject commit_ line. As I said, I have no idea what this is but maybe this info helps?

DaPhil:
> I see that you made it so the WFE is recalculated automatically with a different random instance for each call to get_opd. That's one thing I had thought about doing differently, to support polychromatic calculations in which one would want different wavelengths to all encounter the same phase screens. Perhaps there could be an option to freeze the OPD to a fixed example of the Kolmogorov phase screen to support cases like that. I'll look through the code to see if I have any other comments or suggestions. But I wanted to ask what you thought of this idea.

You could just save the first seed and use it to initialize the phase screens. However, you would need to recalculate all phase screens for all wavelengths which is redundant computing. One could also just save the phase screens for the first wavelength and then use them appropriately for the others (although I am not sure about it, does the pixelscale change with wavelength during propagation? The screens depend on the pixelscale!). But I am not familiar with how you deal with polychromatic waves. If you have them somehow computed in parallel, then at least the second method would not work (and the first still requires a recomputation of all screens although once would be enough).

mperrin:
Closed here in this deprecated repo - see instead  https://github.com/spacetelescope/poppy/pull/437 in the spacetelescope repo (which was merged already)

PR #250, opened 2018-05-15
Unified GPU framework using either CUDA or OpenCL

mperrin:
This branch both adds support for OpenCL and refactors the CUDA implementation somewhat. 

- Essentially all GPU and FFTW related functionality is now relocated into `accel_math.py`, inside wrapper functions, in particular one called `fft_2d`. This allows the algorithmic code in the other files to be written without having to concern itself with _how_ a given FFT is implemented. In principle this extends and generales the `_fft` and `_inv_fft` methods on the FresnelWavefront class (which in fact just become stubs that pass through to `accel_math.fft_2d`. This change involved refactoring key algorithmic parts of both `fresnel.py` and `poppy_core.py`. 
- OpenCL-based implementation of GPU accelerated FFTs added.  Dependencies are pyopencl, conda, gpyfft. The first two can be installed from conda-forge. The latter is in a user contributed conda package, so use ` conda install -c ljbo3 gpyfft`. A new setting `poppy.conf.use_cuda` allows enabling or disabling this code, on machines with OpenCL-compatible GPUs. 
- Work in progress partial implementation of toggling between single and double precision floating point math, controlled by a new setting `poppy.conf.double_precision` which defaults to True. Many things work in single precision mode but currently many test cases fail, presumably since they're not smart enough to adjust tolerances appropriately. Best to ignore this setting for now for most purposes. 

All tests pass locally for me, using any combo of plain numpy, numexpr+fftw, CUDA, or OpenCL. On the other hand this starts to get very machine-dependent in test setups so further testing is much appreciated.

Some areas for future work before merging this:

1. In refactoring the CUDA functionality out of the FresnelWavefront class, that lost the ability to cache CUDA FFT plans as part of FresnelWavefront objects. I tried caching them as a module-level variable in `accel_math.py`, but for unclear reasons doing so led to a memory leak on the GPU (buffers not being freed somehow?), which eventually led to calculations totally failing due to inability to malloc new buffers on the GPU. I couldn't debug where the leak was, but simply _not_ caching the plans avoids it entirely. The time to create a new CUDA FFT plan each time appears < 1 ms, so it's not zero but is relatively negligible. I'm not totally happy with this but reliability comes first. 
2. The single precision stuff is all use-at-your-own-risk. Needs more debugging and attention tothe unit test tolerances.
3. Needs docs! 
4. Travis doesn't support CI for GPUs, so all the meaningful testing for this will need to be run locally. Need to investigate the available CUDA compute servers at STScI. 
5. Benchmark functions would be nice to include. Probably a few relevant cases, perhaps (1) bare FFT (2) coronagraphic multi-plane OpticalSystem calc in Fraunhofer regime (3) Fresnel multi-plane calc.


mperrin:
@douglase here's the PR I promised for you to take a look at. In particular I'm hoping you can confirm that my various edits to your code didn't break anything, or change the performance negatively. (Looks good as far as I can tell, but I'm cautious until I hear it's all working on your hardware as well)

mperrin:
Yeah I realized last night after submitting this that I had accidentally left a bunch of debugging print statements in there unintentionally. Those should mostly just come out entirely. 

I suspect the unnecessary IO is part of what's  causing the slowdowns.

mperrin:
oh, actually it's worse than just unnecessary IO, since those debug statements were also summing the total intensity before and after each FFT. This was leftover from when I was trying to debug some normalization issues.

mperrin:
plus I wasn't using your _fftshift wrapper properly.

poppy/accel_math.py
```diff
@@ -34,19 +31,46 @@
...
+    import gpyfft
+    _OPENCL_AVAILABLE = True
+    _OPENCL_STATE = dict()
+except ImportError:
+    _OPENCL_AVAILABLE = False
+
+
+_USE_CUDA = (conf.use_cuda and _CUDA_AVAILABLE)
+_USE_OPENCL = (conf.use_numexpr and _OPENCL_AVAILABLE)
 _USE_NUMEXPR = (conf.use_numexpr and _NUMEXPR_AVAILABLE)
```
douglase: shouldn't there by a _USE_FFTW for uniformity?

douglase:
a6b632861ad278931c849d343f18e6b743a74139 runtimes are looking good

poppy/accel_math.py
```diff
@@ -64,16 +88,220 @@ def _exp(x):
...
+
+            _OPENCL_STATE['context'] = context
+            _OPENCL_STATE['queue'] = queue
+        return (_OPENCL_STATE['context'], _OPENCL_STATE['queue'])
+
+
+
+
+
 if  _USE_CUDA:
```
douglase: TO

poppy/accel_math.py
douglase: also, what if _USE_CUDA and _USE_OPENCL are both True, is that handled somewhere else?

poppy/accel_math.py
mperrin: Not sure what this comment means?

poppy/accel_math.py
mperrin: It's handled implicitly by the order of the if/elif inside `fit_2d`.  _USE_CUDA has precedence.

douglase: at the moment, I'm flipping these bools to run my benchmarks, that's why I'm paying particular attention to them.

poppy/accel_math.py
douglase: it was supposed to be TO DO: figure out how to rewrite this in OpenCL

poppy/accel_math.py
mperrin: Right. I was trying to get it so that one could flip the `poppy.conf.use_x` bools and it would have the desired effect. On the general principle that what you're trying to do is configure a setting, so it should be possible via the configuration settings system. :-)  

But I see that doesn't quite work right for all the cases of interest right now, since the file global versions are still accessed in a bunch of places.

poppy/accel_math.py
mperrin: Right!  And the lack of that is a big part of why I wrote `fft_2d` to prioritize CUDA rather than OpenCL if both are available...

poppy/accel_math.py
mperrin: well, it only costs a couple microseconds to just evaluate `(conf.use_cuda and _CUDA_AVAILABLE)` on the fly when needed. So there is very small cost to putting that check directly in the fftshift function and so on. That would be one way to let the `conf` settings be toggled at runtime and work.

poppy/accel_math.py
```diff
@@ -64,16 +88,220 @@ def _exp(x):
...
+            # determine optimal algorithm that is destructive to your chosen array.
+            # So only do that test on a copy, not the real array:
+            _log.info("Measuring pyfftw optimal plan for %s, direction=%s" % (
+                str(wavefront.shape), FFT_direction))
+
+            pyfftw.interfaces.cache.enable()
+            pyfftw.interfaces.cache.set_keepalive_time(30)
+
+            test_array = np.zeros(wavefront.shape)
+            test_array = do_fft(test_array, overwrite_input=True, planner_effort='FFTW_MEASURE',
```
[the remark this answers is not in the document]
mperrin: I'm starting to suspect I may have been wrong before about there being a memory leak from storing the plans. It's possible the root cause was actually there's a memory leak _when test cases fail during pytest_, in which case exceptions are raised and variables do not get cleaned up properly.  In which case now that the tests are passing, that should not happen.  

I'm cautiously optimistic this may be the case, and am running the test suite repeatedly on my laptop with CUDA to see...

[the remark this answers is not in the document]
mperrin: I am increasingly convinced this is the case!

mperrin:
I was able to run the test suite a dozen times in a row with no memory errors, which is far beyond the problems I was seeing before. I'm almost entirely convinced the leak was indeed due to failure to cleanup GPU state when calculations error out part way through. Did you encounter anything like that before?  
In any case I re-enabled the plan cache, using the version where it's file global inside `accel_math`, so it should not interfere with Wavefront copying or pickling.

poppy/accel_math.py
```diff
@@ -64,16 +88,213 @@ def _exp(x):
...
+    """
+    # To use a fast FFT, it must both be enabled and the library itself has to be present
+    _USE_CUDA = (conf.use_cuda and _CUDA_AVAILABLE)
+    _USE_OPENCL = (conf.use_opencl and _OPENCL_AVAILABLE)
+    _USE_FFTW = (conf.use_fftw and _FFTW_AVAILABLE)
+
+    # OpenCL cfFFT only can FFT certain array sizes.
+    # This check is more stringent that necessary - opencl can handle powers of a few small integers
+    # but this simple version helps during development
+    if _USE_OPENCL and not ispowerof2(wavefront.shape[0]):
```
mperrin: @douglase are there constraints on which array sizes are allowed for CUDA FFTs? How well does it handle odd sized arrays? Not that I expect those to be constant, but we should check for and avoid cases where it would fail. 

FWIW, the clfft library says it supports "powers of 2, 3, 5, 7, 11 and 13". but here I just check for powers of 2, because that's the common case, and I don't know how to write a computationally-efficient fast check for "factorable into powers of primes no greater than 13".  

I don't necessarily put a high priority on optimal GPU acceleration of weird sized arrays, I just want to make sure the code doesn't encounter any situation that causes a fatal error. Although for some situations it _is_ useful to FFT such arrays; for instance if you need bigger than 4096^2, it's still about 2x faster to FFT an array of size 6144^2 than to jump all the way to 8192^2.

douglase: Ah, this explains why suddenly my odd-sized test was the same size as FFTW. 

CUDA definitely is happy with odd-sizes, but takes a slight hit for unusually sized arrays as seen in this figure from the current master branch where everything is power of two except for a 2800x2800 test:

benchmarks12coresv100

This is consistent the cuFFT documentation (https://docs.nvidia.com/cuda/cufft/index.html#introduction):

'''Features:
* Algorithms highly optimized for input sizes that can be written in the form 2^a×3^b×^5^c^×7^d. In general the smaller the prime factor, the better the performance, i.e., powers of two are fastest.
*An O(nlogn)  algorithm for every input data size
...
'''

douglase: So I would just pass any sized array to CUDA

douglase: hmm, nope, not using opencl and it's only the opencl case your are testing. maybe it's a coincidence of this system. I will keep digging.

douglase: Never mind, it was jus a coincidence of that machine/run.  I am seeing a speed up from CUDA at 2800x2800.
benchmarks12coresv100

poppy/accel_math.py
douglase: It's all a bit confusing but that seems like a reasonable path. I don't think the config file is as critical as having a setting that they can adjust at run time in their own script.

poppy/accel_math.py
mperrin: Can you point me at the benchmark code you’re using for that figure?  You pointed me out a notebook a while ago but I don’t know if that’s still the up-to-date place to find it. I’ll make a similar figure for OpenCL. 

Might want to tweak the X axis label to say “ array dimensions per axis”,  to emphasize that the full array size is the square of that.

douglase: that notebook is still the most current.

mperrin:
@douglase I'm going to go ahead and merge this PR now. I had hoped to get this poppy release done and out the door this week. At this point I'm not expecting to do any additional technical work on this in the near term - need to concentrate efforts on writing up for SPIE etc. 

Incidentally, on my iMac Pro I've ended up in an interesting and surprising situation where _plain numpy_ appears to be outperforming both FFTW and OpenCL. Have tested and benchmarked like crazy since this seemed nuts to me, but it proves to be the case.  Apparently there's major speedups in numpy's intrinsic FFT in the Intel-optimized MKL-linked version of numpy which is now in Conda. This ends up being a very big deal on the iMac Pro since its CPU (Xeon W 2140) supports the AVX-512 SIMD extension, i.e. each core can operate on the equivalent of 4 * complex128s at once.   On this particular hardware that ends up equalling or slightly beating the current result of FFTW (not as well optimized for the SIMD instructions as Intel's hand-tuned code?), and substantially outdoing even this fairly high-end AMD GPU (which is tuned much more for single-precision graphics performance than for double-precision GPGPU work; the chip has many fewer double-precision-capables cores than its total number of cores.)  

Surprising but true, and goes to show that performance optimization has to be tuned for each set of hardware. Totally different optimal paths for my MacBook Pro laptop, iMac Pro desktop, and your EWS-instance GPGPU compute instances... Makes for a more complicated story but appears to be the case.

PR #255, opened 2018-05-31
fix errors statements to refer to the correct calling function name.

douglase:


mperrin:
only changes error message strings - no need to wait for CI to finish.

PR #258, opened 2018-07-02
Fix type of beam_ratio in FresnelOpticalWavefront

sdwill:
The beam_ratio parameter is specified as int, but should be float, since it is given by 1/Q, where Q is the integer entrance pupil padding factor.


mperrin:
Thanks @sdwill, good catch.

PR #263, opened 2019-03-12
First commit for statistical PSD WFE class

ivalaginja:
Please advise if something needs to be different. My units are still off and I need to fix that.


ivalaginja:
I will submit this PR to spacetelecsope/poppy instead.

PR #265, opened 2019-12-07
Thermal Blooming WFE

DaPhil:
Hi there. I implemented a new wavefront error: thermal blooming. If you find it helpful or that it adds value to poppy I am happy to share. I also implemented tests but due to my little experience with testing it might not be complete. There is also a notebook which has an example on how to use it by confirming a result from a paper.


mperrin:
Closed in this deprecated repo - see instead https://github.com/spacetelescope/poppy/pull/438 in the spacetelescope repo

PR #267, opened 2020-01-22
Shack Hartmann WFS class 

remorgan123:
Hi Marshall! 

Here's some code Ewan/I have developed to model a SHWFS through POPPY. It essentially subsamples a wavefront and calculates each spot PSF separately, then tiles it back together to return the full sensor measurement. There are also methods to calculate the centroids of lenslet spots and reconstruct the wavefront as a part of the class. There is a demo notebook in "notebooks/" to show how the class works. Currently, the SHWFS is in the "subapertures.py" file within POPPY, and it works directly on wavefront objects. Since it's a combination of apertures/optical elements/a detector I'm not sure how best to incorporate it into an Optical System class other than retrieving the wavefront at the end of the optical train and measuring it with a SHWFS object. Let me know if you have any suggestions for this or think it would be more appropriate as it's own repository separate from the main POPPY source code!

Another small change is in dms.py - I got confused modeling the DM's in this system since the get_opd function returns the interpolated surface directly. I modified the function to return 2*interpolated_surface since it's a reflective element, but I'm happy to revert this change and multiply by two within the demo script instead if you'd prefer to keep it how it is. It was more convenient for my application to do it this way, but I'm happy to change if other users are used to it how it is, let me know what you think!

Thanks!
- Rachel


remorgan123:
Also - I'm not seeing Ewan as an option of one of the reviewers, is there any way you could add him? @douglase can you see the code this way?

mperrin:
@remorgan123 - thanks, I will take a look at this (though cannot promise to do so in the immediate future). 

However, please note we've moved the main development of `poppy` to the repo in the spacetelescope organization: https://github.com/spacetelescope/poppy 
Could you make a copy of this PR into that repo instead, please? Thanks much.

remorgan123:
Will do, thanks!

poppy/dms.py
```diff
@@ -333,7 +333,7 @@ def get_opd(self, wave):
                     shift_y_pix = int(np.round(self.shift_y /pixscale_m))
                     interpolated_surface = np.roll(interpolated_surface, shift_y_pix, axis=0)
 
-        return interpolated_surface
+        return 2*interpolated_surface # note *2 to account for DM being reflective surface
```
mperrin, line 336: Hi @remorgan123 - I would like to _not_ include this factor of 2 here, or at least to make it optional via some class parameter. The implicit convention thus far in poppy has been to have the DM classes modeled in terms of OPD rather than physical surface (despite the variable names...).  I realize this is potentially confusing and hasn't been super well documented though.  More significantly, changing this behavior now would be introduce unexpected factor of 2 change into any existing code that uses the poppy DM classes; i.e. this is not a back-compatible change. 

Instead I think we could have this as a new parameter to the class init, e.g. `include_factor_of_two=False` or similar. And then here would have more like: 
```
coefficient = 2 if self.include_factor_of_two else 1
return coeff*interpolated_surface
```

Doing it that way would let us gradually shift the behavior here, without immediately introducing unexpected factor of 2 change for all existing uses of this code...

mperrin (changes requested):
Hi @remorgan123 - Thanks much. A few more suggestions & requests:

- I see this currently contains both "SHWFS Demo.ipynb" and "Shack Hartmann Wavefront Sensor Demo.ipynb". It looks like that's redundant and the "SHWFS" one should be deleted - is that correct? 

- When running the example notebook, in cell 6 with the zonal reconstruction it outputs some errors "Algorithm terminated at max iterations without convergence.". Is that expected, and if so why? 

- I already commented separately that I would like to make the factor of 2 change for surface-to-OPD in the DM code optional, to avoid introducing that change into existing working application code that doesn't expect it. 

- And lastly, here's a suggestion for the last cell of the example notebook. We can adjust the plot parameters somewhat and also include a comparison plot of the input wavefront, to show that the reconstructed wavefront is correctly sampling and reconstructing it. Try like this: 

```python
#display result:
plt.figure()
plt.imshow(reconstruction.value, cmap=poppy.conf.cmap_diverging, vmin=-1, vmax=1)
plt.colorbar(label = "Wavefront error, " + str(reconstruction.unit))
plt.xlabel("x lenslet")
plt.ylabel("y lenslet")
plt.title("Reconstructed Wavefront")


plt.figure()
dm.display(opd_vmax=1e-6, 
           colorbar_orientation='vertical', 
           title='Input Wavefront (Truth)')
```

mperrin:
Whoops - added PR comments on the wrong instance of this. Going to move over to https://github.com/spacetelescope/poppy/pull/346 instead.

mperrin:
@remorgan123 I'm going to go ahead and close this PR here, and we'll just work on the https://github.com/spacetelescope/poppy/pull/346 one from here out.

douglase:
I agree, the default behavior should be unity, and i might only add the
coefficient option if it gets added to the more general optical surfaces
class consistently.

PR #270, opened 2021-02-26
Updates to fresnel.py allowing users to create their own inwave object.

kian1377:
The changes made are such that a user can input their own custom wavefront data into a FresnelOpticalSystem() when using calc_psf(). 

The methods calc_psf() and propagate_mono() have been added directly into the FresnelOpticalSystem() class, so they are not just inherited and they contain an additional kwarg that is inwave. This kwarg is initialized to be None, but the user can define their own FrenselWavefront() and input it into calc_psf() using the inwave kwarg. This inwave object is then passed to propagate_mono(), which subsequently passes inwave to the input_wavefront() method in FresnelOpticalSystem(), which has also been edited to initialize a wavefront if the value of inwave is None, or to raise a value error if something other than a FresnelWavefront() object was supplied. 

The changes made have been in an effort to create the Roman CGI modes with POPPY, however, the SPC modes and their data files required a unique sequence of MFTs and FFTs to apply the Focal Plane Masks (FPMs). So two separate optical systems were created, the first going up to the FPM and the second going through the rest of the optics of the CGI. This way, the FPM could be applied to the final wavefront of the first optical system, and then input into the second system to continue propagation. The results of this system have been tested and show agreement with the results of the PROPER model. 

image

Future efforts will include implementing the technique used to apply the FPMs directly into POPPY such that two separate optical systems do not need to be created, but having the flexibility for a user to input their own wavefront data still seemed useful. 

Please let me know if there issues with the PR or anything else I should change. Others who have contributed are @douglase and @Jashcraf.

### DIFF
--- a/poppy/fresnel.py
+++ b/poppy/fresnel.py
@@ -12,8 +12,10 @@ if accel_math._USE_NUMEXPR:
     import numexpr as ne
     pi = np.pi  # needed for evaluation inside numexpr strings.
 
-_log = logging.getLogger('poppy')
+from .accel_math import _float, _complex
+from . import conf
 
+_log = logging.getLogger('poppy')
 
 __all__ = ['QuadPhase', 'QuadraticLens', 'FresnelWavefront', 'FresnelOpticalSystem']
 
@@ -75,8 +77,6 @@ class QuadPhase(poppy.optics.AnalyticOpticalElement):
             opd = (x ** 2 + y ** 2)  / (2.0 *z)
 
         return opd
-
-
 
 class _QuadPhaseShifted(QuadPhase):
     """
@@ -168,7 +168,7 @@ class ConicLens(poppy.optics.CircularAperture):
 class FresnelWavefront(BaseWavefront):
     angular_coordinates = False
     """Should coordinates be expressed in arcseconds instead of meters at the current plane? """
-
+    
     @u.quantity_input(beam_radius=u.m)
     def __init__(self,
                  beam_radius,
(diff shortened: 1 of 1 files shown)


mperrin:
@kian1377 - Thanks!  One request please, we're now doing development of poppy in the spacetelescope repo: https://github.com/spacetelescope/poppy Can you create a copy of this pull request there instead, please?  Much appreciated. 

I won't be able to look at this in depth until next week or later. But you might also want to take a look at the CompoundOpticalSystem class. See https://github.com/spacetelescope/poppy/blob/develop/poppy/poppy_core.py#L2124

I developed this last year in order to handle a system model that mixes Fresnel and Fraunhofer propagations, with the output from one going as input to the next.   That's another way to take your two separate optical systems and chain them together.  This is not well documented yet - it's something I put together to support optical modeling for the HICAT lab testbed here, but it hasn't made it to the top of the to-do list yet to add documentation beyond just the doc string of that function. 

(And very nice to see those simulations of the CGI SPC!)

kian1377:
Thanks for getting back so quickly, I will move this PR over to the spacetelescope branch of POPPY. 

We took a look at using CompoundOpticalSystem() as well, but our specific application wasn't converting between wavefront or optical system types, the issue was simply the procedure used to apply the FPM in the PROPER models because the MFT and FFT sequence allows for high sampling at the FPM and does not require knowing the FPM pixelscale in units of meters/pixels, just in lambda/D units. I am looking into adding this method of applying a FPM into POPPY by defining a new method in the FresnelWavefront() class which is used during propagation and only called if a specific kwarg is provided at the FPM plane of the system in order to actually apply the FPM to the wavefront that has been propagated to that plane. 

As for the PR itself, it seems the Travis build has failed so I will look into why it is failing as well.

mperrin:
Thanks - and I’d say hold off on looking at Travis until seeing if this passes on the main repo

mperrin:
Closed here in this deprecated repo - see instead https://github.com/spacetelescope/poppy/pull/423 in the new repo

